PR #30, opened 2021-01-18
AWS driver supports other protocols

congoamz:
### Summary

AWS driver supports other protocols

### Description

- AWS driver supports other protocols
- corrected license value to GPLv2
- logger is off by default

### Related Issue

RDS-242

### Additional Reviewers

@hsuamz 
@seneramz 
@sergiyvamz


src/test/java/testsuite/regression/ConnectionRegressionTest.java
```diff
@@ -11565,7 +11520,7 @@ public void testBug98445() throws Exception {
         System.setErr(new PrintStream(newErr));
 
         // clientInfoProvider=CommentClientInfoProvider
-        testConn1 = getConnectionWithProps("clientInfoProvider=CommentClientInfoProvider,profileSQL=true");
+        testConn1 = getConnectionWithProps("clientInfoProvider=CommentClientInfoProvider,profileSQL=true,logger=StandardLogger");
```
sergiyv-bitquill, line 11523: why do we need to setup a looger?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -276,9 +274,10 @@ protected synchronized void initSettings(ConnectionUrl connectionUrl) throws SQL
     this.failoverSocketTimeoutMs = socketTimeout.isExplicitlySet() ? socketTimeout.getValue() : DEFAULT_SOCKET_TIMEOUT_MS;
   }
 
-  protected synchronized void initLogger() {
-    if (!StringUtils.isNullOrEmpty(this.loggerClassNameSetting)) {
-      this.log = LogFactory.getLogger(this.loggerClassNameSetting, Log.LOGGER_INSTANCE_NAME);
+  protected synchronized void initLogger(ConnectionUrl connUrl) {
```
sergiyv-bitquill, line 277: I expected just a simple change in PropertyDefinitions that change default logger class.

aaron-congo: Hmm I think I mentioned that if we change the default logger class to null_logger in property definitions, then logging will be disabled in other parts of the code external to our driver. Since those parts of the code use the standard logger by default right now, would it be best to avoid changing that behavior? Maybe I'm still confused about this though

src/test/java/testsuite/regression/ConnectionRegressionTest.java
aaron-congo: this test uses the aws protocol, and with this pr logging is off by default for the aws protocol. As far as I can tell the test relies on logging information being spit out (see the assertion on line 11532). This assertion started failing when I turned the logging off and passes when I add this connection property to turn it back on

sergiyvamz (approved):

PR #41, opened 2021-03-11
Bug Fix: Use Current Connection Properties After Failover

congoamz:
### Summary

Bug Fix: Use Current Connection Properties After Failover

### Description

- Fixed a bug where cached properties from previous connections were used after failover instead of the properties configured for the current connection
- Failover integration test 5.1 now includes an initial step that establishes the topology cache to test this fix
- Includes some minor changes suggested by intellij

### Additional Reviewers

@sergiyvamz 
@seneramz 
@hsuamz


src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareReaderFailoverHandler.java
```diff
@@ -343,16 +346,19 @@ private ConnectionAttemptResult getNextResult(CompletionService<ConnectionAttemp
     private final ConnectionProvider connProvider;
```
congoamz, line 346: should we make this class non-static so that we don't have to pass in so many parameters to the constructor? Might be cleaner

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
```diff
@@ -67,6 +69,7 @@
   protected int maxFailoverTimeoutMs = 60000; // 60 sec
```
congoamz, line 69: Similarly to ConnectionAttemptTask in ClusterAwareReaderFailover, should we make the two static classes in this class non-static so that we don't have to pass in so many parameters to the constructor?

src/main/core-api/java/com/mysql/cj/util/ClusterAwareUtils.java
```diff
@@ -0,0 +1,30 @@
...
+     * Create a copy of the given {@link HostInfo} object where all details are the same except for the host properties,
+     * which will contain both the original properties and the properties passed into the function
+     *
+     * @param baseHostInfo The {@link HostInfo} object to copy
+     * @param newProps The map of properties to add to the new {@link HostInfo} copy
+     *
+     * @return A copy of the given {@link HostInfo} object where all details are the same except for the host properties,
+     *      will contain both the original properties and the properties passed into the function
+     */
+    public static HostInfo hostInfoCopyWithAdditionalProps(HostInfo baseHostInfo, Map<String, String> newProps) {
```
sergiyvamz, line 22: just copy() maybe?

src/main/core-api/java/com/mysql/cj/util/ClusterAwareUtils.java
```diff
@@ -0,0 +1,30 @@
+package com.mysql.cj.util;
```
sergiyvamz, line 1: com.mysql.cj.jdbc.ha.ca ?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareHitMissMetricsHolder.java
```diff
@@ -80,10 +80,10 @@ public void reportMetrics(Log log) {
     logMessage.append("** Performance Metrics Report for '");
     logMessage.append(this.metricName);
     logMessage.append("' **\n");
-    logMessage.append("\nNumber of reports: " + this.numberOfReports);
```
sergiyvamz, line 83: That "+" makes extra internal copies of strings. It's better to use two separate statements like this:
logMessage.append("\nNumber of reports: ");
logMessage.append(this.numberOfReports);

It's about performance.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareReaderFailoverHandler.java
sergiyvamz: Sure, I'd keep just newHostTuple as a single constructor parameter then.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
sergiyvamz: Yes, it's a good idea.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareHitMissMetricsHolder.java
aaron-congo: yeah I think that's what this change is doing right? The + is being removed in this commit

sergiyvamz (approved):

PR #42, opened 2021-03-16
Add setAwsProtocolOnly switch to mysql jdbc

sertay-s:
### Summary

Add setAwsProtocolOnly switch.

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/NonRegisteringDriver.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/NonRegisteringDriver.java
@@ -75,6 +75,8 @@ import com.mysql.cj.util.StringUtils;
  */
 public class NonRegisteringDriver implements java.sql.Driver {
 
+    protected static boolean acceptAwsProtocolOnly = false;
+
     /*
      * Standardizes OS name information to align with other drivers/clients
      * for MySQL connection attributes
@@ -196,7 +198,8 @@ public class NonRegisteringDriver implements java.sql.Driver {
 
             ConnectionUrl conStr = ConnectionUrl.getConnectionUrlInstance(url, info);
             Map<String, String> connProps = conStr.getOriginalProperties();
-            boolean acceptAwsProtocolOnly = Boolean.parseBoolean(connProps.getOrDefault(PropertyKey.acceptAwsProtocolOnly.getKeyName(), "false"));
+            acceptAwsProtocolOnly = Boolean.parseBoolean(connProps.getOrDefault(PropertyKey.acceptAwsProtocolOnly.getKeyName(), "false"));
+
             switch (conStr.getType()) {
                 case SINGLE_CONNECTION:
                     return acceptAwsProtocolOnly ? null : com.mysql.cj.jdbc.ConnectionImpl.getInstance(conStr.getMainHost());
--- a/src/test/java/com/mysql/cj/jdbc/NonRegisteringDriverTest.java
+++ b/src/test/java/com/mysql/cj/jdbc/NonRegisteringDriverTest.java
@@ -37,16 +37,36 @@ import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
 
+import java.lang.reflect.Method;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Properties;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.*;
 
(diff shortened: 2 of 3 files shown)


congoamz:
Can we change the row for this connection parameter in the README to the following:
|`acceptAwsProtocolOnly` | Boolean | If using simultaneously with another postgresql driver that supports the same protocols, it is required to set this property to true either through the connection properties or at the driver level for every connection by calling Driver.setAcceptAwsProtocolOnly(true). <br/>Otherwise: No | Set to true to only accept connections for URLs with the jdbc:postgresql:aws:// protocol. This setting should be set to true when running an application that uses this driver simultaneously with another postgresql driver that supports the same protocols (eg the PostgreSQL JDBC Driver), to ensure the driver protocols do not clash. This behavior can also be set at the driver level for every connection via the Driver.setAcceptAwsProtocolOnly method; however, this connection parameter will take priority when present<br/><br/>**Default value:** `false`

src/main/user-impl/java/com/mysql/cj/jdbc/NonRegisteringDriver.java
```diff
@@ -75,6 +75,11 @@
  */
 public class NonRegisteringDriver implements java.sql.Driver {
 
+    protected static boolean acceptAwsProtocolOnly = false;
+
+    private static final String MYSQL_PROTOCOL = "jdbc:mysql:";
+    private static final String AWS_PROTOCOL = MYSQL_PROTOCOL + "aws:";
```
congoamz, line 81: probably can remove these two and use SINGLE_CONNECTION_AWS.getScheme() instead of AWS_PROTOCOL

src/main/user-impl/java/com/mysql/cj/jdbc/NonRegisteringDriver.java
```diff
@@ -196,7 +204,8 @@ public boolean acceptsURL(String url) throws SQLException {
 
             ConnectionUrl conStr = ConnectionUrl.getConnectionUrlInstance(url, info);
             Map<String, String> connProps = conStr.getOriginalProperties();
-            boolean acceptAwsProtocolOnly = Boolean.parseBoolean(connProps.getOrDefault(PropertyKey.acceptAwsProtocolOnly.getKeyName(), "false"));
+            acceptAwsProtocolOnly = Boolean.parseBoolean(connProps.getOrDefault(PropertyKey.acceptAwsProtocolOnly.getKeyName(), "false"));
```
congoamz, line 207: I think there are a couple potential problems here. Seems like there's two levels at which we can set this attribute: the connection level for just this specific connection or the driver level for every single connection. I think if we set it as a connection parameter it shouldn't modify the driver-level attribute. Also, instead of defaulting to "false", shouldn't we default to the driver-level value? We want to use the connection parameter when present, but if it isn't, then use the driver-level value

src/test/java/com/mysql/cj/jdbc/NonRegisteringDriverTest.java
```diff
@@ -37,16 +37,36 @@
...
+import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.*;
 
 @SuppressWarnings("unchecked")
 public class NonRegisteringDriverTest {
+
+    @Test
+    public void testSetAwsProtocolOnlySwitch() throws Exception {
```
congoamz, line 57: I think this test needs a few more cases to cover the different combinations of driver-level and connection-level settings. I'll follow up directly on slack with you about it

src/main/user-impl/java/com/mysql/cj/jdbc/NonRegisteringDriver.java
seneramz: We might have a different understanding of the use case of these switches, following up offline.

seneramz: I think our current understanding is that we are providing two different ways to set the same Driver level variable. If this changes that part should be modified as discussed.

seneramz: It changed and that part is adjusted :)

src/test/java/com/mysql/cj/jdbc/NonRegisteringDriverTest.java
```diff
@@ -37,16 +37,45 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.*;
 
 @SuppressWarnings("unchecked")
 public class NonRegisteringDriverTest {
+
```
congoamz, line 55: should we maybe add a BeforeEach here that calls Driver.setAcceptAwsProtocolOnly(false) in case we add more tests later or change this one and don't want the static setting to carry over to other tests?

sergiyvamz (approved):

PR #53, opened 2021-04-01
Improve integration tests timeouts

sergiyvamz:
### Summary

Improve integration tests timeouts

### Description

Use sockets blocking communication to db cluster instances instead of repeatedly running fault injection SQL.

### Related Issue

RDS-381

### Additional Reviewers

@congoamz 
@samathaamz 
@seneramz 
@hsuamz

### DIFF
--- a/src/test/java/testsuite/failover/FailoverIntegrationTest.java
+++ b/src/test/java/testsuite/failover/FailoverIntegrationTest.java
@@ -30,6 +30,8 @@
 
 package testsuite.failover;
 
+import testsuite.UnreliableSocketFactory;
+
 import com.amazonaws.services.rds.AmazonRDS;
 import com.amazonaws.services.rds.AmazonRDSClientBuilder;
 import com.amazonaws.services.rds.model.DBCluster;
@@ -55,10 +57,14 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.HashMap;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -71,21 +77,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 @Disabled
 @TestMethodOrder(MethodOrderer.Alphanumeric.class)
 public class FailoverIntegrationTest {
-  /*
-   * Before running these tests we need to initialize the test cluster as the following.
-   *
-   * Expected cluster state:
-   * +------------------+--------+---------+
-   * |   Instance Id    |  Role  | Status  |
-   * +------------------+--------+---------+
-   * | mysql-instance-1 | Writer | Running |
-   * | mysql-instance-2 | Reader | Running |
-   * | mysql-instance-3 | Reader | Running |
-   * | mysql-instance-4 | Reader | Running |
-   * | mysql-instance-5 | Reader | Running |
-   * +------------------+--------+---------+
-   *
-   * */
+
   private static final String DB_CONN_STR_PREFIX = "jdbc:mysql:aws://";
   private static final String DB_CONN_STR_SUFFIX = System.getenv("DB_CONN_STR_SUFFIX");
   private static final String DB_READONLY_CONN_STR_SUFFIX =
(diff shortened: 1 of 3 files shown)


src/test/java/testsuite/failover/FailoverIntegrationTest.java
```diff
@@ -193,18 +177,24 @@ public void test1_2_failFromWriterToNewWriter_failOnConnectionBoundObjectInvocat
...
+    props.setProperty(PropertyKey.socketFactory.getKeyName(), testsuite.UnreliableSocketFactory.class.getName());
+    props.setProperty(PropertyKey.failoverTimeoutMs.getKeyName(), "10000");
+    testConnection = connectToWriterInstance(initalWriterId, props);
 
     // Crash all reader instances (2 - 5).
-    startCrashingInstanceAndWaitUntilDown(INSTANCE_ID_2);
-    startCrashingInstanceAndWaitUntilDown(INSTANCE_ID_3);
-    startCrashingInstanceAndWaitUntilDown(INSTANCE_ID_4);
-    startCrashingInstanceAndWaitUntilDown(INSTANCE_ID_5);
+    for (int i = 2; i < instanceIDs.length; i++) {
```
congoamz, line 192: should this start from i = 1? The test originally crashes all but the first instance I think

src/test/java/testsuite/failover/FailoverIntegrationTest.java
```diff
@@ -610,72 +650,47 @@ public void test3_4_writerFailWithNoTransaction() throws SQLException, Interrupt
...
     final String nextWriterId = getDBClusterWriterInstanceId();
     assertEquals(nextWriterId, currentConnectionId);
-    assertEquals(INSTANCE_ID_2, currentConnectionId);
+    assertEquals(instanceIDs[1], currentConnectionId);
 
     // Assert that the pooled connection is valid.
     assertTrue(testConnection.isValid(IS_VALID_TIMEOUT));
-
-    // Start crashing all other instances
-    startCrashingInstanceAndWaitUntilDown(INSTANCE_ID_3);
```
congoamz, line 634: are these chunks just not necessary anymore?

src/test/java/testsuite/failover/FailoverIntegrationTest.java
```diff
@@ -928,140 +984,129 @@ private void assertFirstQueryThrows(Statement stmt, String expectedSQLErrorCode)
...
+  protected void makeSureInstancesUp(String[] instances, boolean finalCheck) throws InterruptedException {
+    this.log.logDebug("Wait until the following instances are up: \n" + String.join("\n", instances));
+    ExecutorService executorService = Executors.newFixedThreadPool(instances.length);
+    final HashSet<String> remainingInstances = new HashSet<String>();
+    remainingInstances.addAll(Arrays.asList(instances));
+
+    for (String id : instances) {
+      executorService.submit(() -> {
+        while (true) {
+          try (Connection conn = createCrashConnection(id)) {
```
congoamz, line 1055: should we rename createCrashConnection because the implementation of the method itself doesn't have to do with crashing? And it seems like this is replacing createConnectionWithProxyDisabled; looks like createConnectionWithProxyDisabled isn't being called anymore. Maybe instead of creating this new method, just update the createConnectionWithProxyDisabled implementation with the createCrashConnection implementation?

src/test/java/testsuite/failover/FailoverIntegrationTest.java
```diff
@@ -928,140 +984,129 @@ private void assertFirstQueryThrows(Statement stmt, String expectedSQLErrorCode)
...
-    @Override
-    public void run() {
-      while (keepCrashingInstance) {
-        try (Connection conn = createConnectionWithProxyDisabled(instanceId);
-             Statement myStmt = conn.createStatement()
-        ) {
-          myStmt.execute(CRASH_QUERY);
-        } catch (SQLException e) {
-          // Do nothing and keep creating connection to crash instance.
+  protected void makeSureInstancesDown(String[] instances, boolean finalCheck) throws InterruptedException {
```
congoamz, line 1082: Looks like finalCheck is always true, so we can probably get rid of the makeSureInstancesDown methods above and remove the finalCheck boolean parameter

src/test/java/testsuite/failover/FailoverIntegrationTest.java
sergiyvamz: No, they're replaced by another set of methods that crash/check nodes in parallel.

src/test/java/testsuite/failover/FailoverIntegrationTest.java
sergiyvamz: Good catch. Fixed.

src/test/java/testsuite/failover/FailoverIntegrationTest.java
congoamz: Right but is the assertFalse(testConnection.isValid(IS_VALID_TIMEOUT)) important to test still or no?

src/test/java/testsuite/failover/FailoverIntegrationTest.java
sergiyvamz: What'd be a better name for createCrashConnection ?

sergiyvamz: Sure, I removed createCrashConnection and use createConnectionWithProxyDisabled

src/test/java/testsuite/failover/FailoverIntegrationTest.java
sergiyvamz: reviveInstancesAndCloseTestConnection(), that executes after each test, uses finalCheck=false. Test shouldn't fail if not all instances get back online after test.

congoamz: looks like that method uses finalCheck=false for makeSureInstancesUp, I'm talking about makeSureInstancesDown here though

src/test/java/testsuite/failover/FailoverIntegrationTest.java
congoamz: followed up offline, wasn't necessary

congoamz (approved):

PR #55, opened 2021-04-08
Handle topology queries that don't return a writer

congoamz:
### Summary

Handle topology queries that don't return a writer

### Description

- also added guards against exceptions that might arise from calling this.hosts.get when this.hosts is null or empty, or this.hosts.get(this.currentHostIndex) when this.currentHostIndex == -1
- reader failover, writer failover, ClusterAwareReaderFailover#getReaderConnection, and AuroraTopologyService#getTopology never return null

### Additional Reviewers

@seneramz 
@sergiyv-bitquill 
@hsuamz 
@samathaamz

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
@@ -680,12 +680,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
       return;
     }
 
-    if(Util.isNullOrEmpty(this.hosts)) {
-      this.log.logDebug("ClusterAwareConnectionProxy.21");
-      return;
-    }
-
-    if (isConnected()) {
+    if (isConnected() || Util.isNullOrEmpty(this.hosts)) {
       failover(this.currentHostIndex);
       return;
     }
@@ -902,7 +897,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
     this.log.logDebug(Messages.getString("ClusterAwareConnectionProxy.17"));
 
     HostInfo failedHost = null;
-    if(failedHostIdx != NO_CONNECTION_INDEX && !Util.isNullOrEmpty(this.hosts)) {
+    if (failedHostIdx != NO_CONNECTION_INDEX && !Util.isNullOrEmpty(this.hosts)) {
       failedHost = this.hosts.get(failedHostIdx);
     }
     ReaderFailoverResult result = readerFailoverHandler.failover(this.hosts, failedHost);
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
@@ -569,6 +569,10 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
   }
 
   private int getCandidateIndexForInitialConnection() {
+    if(Util.isNullOrEmpty(this.hosts)) {
+      return NO_CONNECTION_INDEX;
+    }
+
     if (isExplicitlyReadOnly()) {
       int candidateReaderIndex = getCandidateReaderForInitialConnection();
       if (candidateReaderIndex != NO_CONNECTION_INDEX) {
@@ -724,7 +728,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
                 new StringBuilder("Connection to ")
                         .append(isWriterHostIndex(hostIndex) ? "writer" : "reader")
                         .append(" host '")
-                        .append(host.getHostPortPair())
+                        .append(host == null ? "<null>" : host.getHostPortPair())
                         .append("' failed");
         try {
           this.log.logWarn(msg.toString(), e);
@@ -859,7 +863,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
       this.failoverStartTimeMs = 0;
     }
 
-    if (!failoverResult.isConnected()) {
+    if (failoverResult == null || !failoverResult.isConnected()) {
       // "Unable to establish SQL connection to writer node"
       processFailoverFailure(Messages.getString("ClusterAwareConnectionProxy.2"));
       return;
@@ -908,7 +912,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
       this.failoverStartTimeMs = 0;
     }
 
-    if (!result.isConnected()) {
+    if (result == null || !result.isConnected()) {
       // "Unable to establish SQL connection to reader node"
       processFailoverFailure(Messages.getString("ClusterAwareConnectionProxy.4"));
       return;
@@ -1320,7 +1324,7 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
       msg.append("\n   [")
               .append(i)
               .append("]: ")
-              .append(hostInfo.getHost());
+              .append(hostInfo == null ? "<null>" : hostInfo.getHost());
     }
     this.log.logTrace(
         Messages.getString("ClusterAwareConnectionProxy.11", new Object[] {msg.toString()}));
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
@@ -930,13 +930,11 @@ public class ClusterAwareConnectionProxy extends MultiHostConnectionProxy
 
     if(this.currentHostIndex != NO_CONNECTION_INDEX && !Util.isNullOrEmpty(this.hosts)) {
       HostInfo currentHost = this.hosts.get(this.currentHostIndex);
+      topologyService.setLastUsedReaderHost(currentHost);
       this.log.logDebug(
               Messages.getString(
                       "ClusterAwareConnectionProxy.15",
                       new Object[] {currentHost}));
-      if (currentHost != null) {
-        topologyService.setLastUsedReaderHost(currentHost);
-      }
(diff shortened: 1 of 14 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -568,7 +575,7 @@ private int getCandidateIndexForInitialConnection() {
         return candidateReaderIndex;
       }
     }
-    return this.hosts.get(WRITER_CONNECTION_INDEX) != null ? WRITER_CONNECTION_INDEX : NO_CONNECTION_INDEX;
+    return WRITER_CONNECTION_INDEX;
```
sergiyvamz, line 578: Don't we need to return NO_CONNECTION_INDEX in case topology is empty?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -674,6 +680,11 @@ protected synchronized void pickNewConnection() throws SQLException {
       return;
     }
 
+    if(Util.isNullOrEmpty(this.hosts)) {
```
sergiyvamz, line 683: This block prevents calling failover() few line below. Failover() may update topology and successfully connect.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -723,7 +729,7 @@ private synchronized void connectTo(int hostIndex) throws SQLException {
                 new StringBuilder("Connection to ")
                         .append(isWriterHostIndex(hostIndex) ? "writer" : "reader")
                         .append(" host '")
-                        .append(host == null ? "<null>" : host.getHostPortPair())
+                        .append(host.getHostPortPair())
```
sergiyvamz, line 732: No harm to leave a safer statement.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -858,13 +864,13 @@ protected void failoverWriter() throws SQLException {
       this.failoverStartTimeMs = 0;
     }
 
-    if (failoverResult == null || !failoverResult.isConnected()) {
+    if (!failoverResult.isConnected()) {
```
sergiyvamz, line 867: I'd keep previous code. It's safer.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -905,9 +913,10 @@ protected void failoverReader(int failedHostIdx) throws SQLException {
       this.failoverStartTimeMs = 0;
     }
 
-    if (result == null || !result.isConnected()) {
+    if (!result.isConnected()) {
```
sergiyvamz, line 916: I'd keep the previous code. It's safer.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
```diff
@@ -1318,7 +1325,7 @@ private void logTopology() {
       msg.append("\n   [")
               .append(i)
               .append("]: ")
-              .append(hostInfo == null ? "<null>" : hostInfo.getHost());
+              .append(hostInfo.getHost());
```
sergiyvamz, line 1328: The previous code is safer.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
```diff
@@ -188,11 +182,12 @@ private WriterFailoverResult getNextResult(ExecutorService executorService,
 
   private void logTaskSuccess(WriterFailoverResult result) {
     List<HostInfo> topology = result.getTopology();
-    if(topology == null || topology.isEmpty() || topology.get(WRITER_CONNECTION_INDEX) == null) {
+    if(Util.isNullOrEmpty(topology)) {
       String taskId = result.isNewHost() ? "TaskB" : "TaskA";
```
sergiyvamz, line 186: That's black magic.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
```diff
@@ -431,7 +422,7 @@ private void logTopology() {
         msg.append("\n   [")
             .append(i)
             .append("]: ")
-            .append(hostInfo == null ? "<null>" : hostInfo.getHost());
+            .append(hostInfo.getHost());
```
sergiyvamz, line 425: The previous code is safer.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: The only function that calls this one has a few steps that it takes before calling this function, and among those steps it verifies that the topology is not null or empty. So this function is not being called until after that has already been verified. However I'll add another check here in case we add more code that calls this function in the future

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: If the host list is empty I don't think we want to fail over, we don't have any host/topology information so I don't think failover is possible (both failover procedures rely on being able to pull host information from the topology). And I don't think we can update topology because we don't use the current connection and we have no other information for hosts to connect to. Correct me if I'm wrong though

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: The only way the host would be null at this point is if we had a null entry in the topology list, which won't happen as far as I can tell since we removed the null writer implementation. However if you think I should add it back still I can do that

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: The only way the host would be null at this point is if we had a null entry in the topology list, which won't happen as far as I can tell since we removed the null writer implementation. However if you think I should add it back still I can do that

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
congoamz: The only way the host would be null at this point is if we had a null entry in the topology list, which won't happen as far as I can tell since we removed the null writer implementation. However if you think I should add it back still I can do that

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
congoamz: haha what do you mean?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: both the failover procedures don't return null anymore; they always return a failed or successful result. But if you want me to add the check back in case we change the implementation in the future then I can add it back

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
congoamz: both the failover procedures don't return null anymore; they always return a failed or successful result. But if you want me to add the check back in case we change the implementation in the future then I can add it back

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
sergiyvamz: I see your point. I just prefer to be on safe side.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareConnectionProxy.java
sergiyvamz: I'd prefer it back.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ca/ClusterAwareWriterFailoverHandler.java
sergiyvamz: Yes, please use previous code.

src/test/java/testsuite/failover/FailoverIntegrationTest.java
```diff
@@ -164,7 +164,7 @@ public void test1_2_failFromWriterToNewWriter_failOnConnectionBoundObjectInvocat
     failoverClusterAndWaitUntilWriterChanged(initialWriterId);
 
     // Failure occurs on Statement invocation
-    assertFirstQueryThrows(stmt, "08S02");
```
sergiyvamz, line 167: I'd keep original implementation since it's more flexible.

sergiyvamz (approved):

PR #98, opened 2021-09-20
fix: Fix source and target Java level not set to 8

karenc-bq:
### Summary

Fix source and target Java level not set to 8

### Description

- Set the `sourceCompatibility` and the `targetCompatibility` values to Java 1.8 so project compiles to Java 8.
- Remove `you can also use 'software.aws.rds.jdbc.mysql.Driver()' to register`, this line is only printed when the driver has been registered.
- Remove `implements Driver` since `NonRegisteringDriver` already implements it.

### Additional Reviewers

@sergiyv-bitquill 
@hsuamz 
@congoamz

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,6 +53,8 @@ plugins {
 java {
     withJavadocJar()
     withSourcesJar()
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 }
 
 checkstyle {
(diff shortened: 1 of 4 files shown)


congoamz (approved):

build.gradle.kts
```diff
@@ -112,14 +114,6 @@ tasks.shadowJar {
         into("META-INF/services/")
     }
 
-    doFirst {
```
sergiyv-bitquill, line 115: What's the purpose of removing this code block? That's about service provider support.

karenc-bq: We can include the `META-INF/services/java.sql.Driver` file directly in the project, i.e. `src/main/resources/META-INF/services/java.sql.Driver`, this allows for:
1. Avoids the need to call `Class.forName(software.aws.rds.jdbc.mysql.Driver)` when writing tests utilizing `DriverManager.getConnection()`
2. Removes the need to create a new file every time we build the JAR

sergiyv-bitquill (approved):

PR #119, opened 2021-11-01
Merge main

karenc-bq:
### Summary

Ensure main branches are up to date.

### Additional Reviewers
@hsuamz 
@ColinKYuen


src/test/java/testsuite/integration/AwsIamAuthenticationIntegrationTest.java
```diff
@@ -38,7 +38,7 @@
 import java.sql.SQLException;
 import java.util.Properties;
 
-@Disabled
+//@Disabled
```
justing-bq: Why not just removed this `@Disabled` flag instead of commenting it out?

karenc-bq: We should disable this test.

ColinKYuen (approved):

PR #174, opened 2022-02-22
feat: Failover Connection Plugin

ColinKYuen:
### Summary

Failover Connection Plugin

### Description

- Failover functionality refactored into a connection plugin.
- Upstream changes from MySQL 8.0.28 community driver.
- Docker containers are now created as part of testing and used to run both unit and integration tests.
- Failover performance metrics tracked based on a cluster. With options to enable additional performance metrics per instance.
- Fixed issue in Enhanced Failure Monitoring connection status check.

### Additional Reviewers

@karenc-bq @hsuamz @sergiyvamz


sergiyv-bitquill (approved):

sergiyvamz (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  @Override
+  public synchronized Object invoke(Object proxy, Method method, Object[] args)
+      throws Throwable {
+    final String methodName = method.getName();
+
+    if (isDirectExecute(methodName, args)) {
+      return executeMethodDirectly(methodName, args);
+    }
```
buitcj, line 183: Hey @ColinKYuen I am confused about this section of the code. Can you help me understand it? There are ClientPreparedStatements created with this ConnectionProxy object. If we execute hashCode() and equals() directly on the proxy object, it would mean that multiple statments created by the same ConnectionProxy would collid in java hash collections. Is this expected? Hibernate is registering my ClientPreparedStatements but it is failing due to finding that a newly registered statement already exists in its internal set even though the already-registered statement is not the same. It only appears to be the same since equals and hashCode just use the same ConnectionProxy. I think this is a bug.

PR #201, opened 2022-04-22
increment subminor version for snapshot builds

sergiyvamz:
### Summary

Automatically increment patch version for snapshot builds

### Additional Reviewers

@hsuamz 
@ColinKYuen

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,7 +33,7 @@ import org.apache.tools.ant.filters.ReplaceTokens
 // Driver version numbers
 val versionMajor = project.property("com.mysql.cj.build.driver.version.major")
 val versionMinor = project.property("com.mysql.cj.build.driver.version.minor")
-val versionSubminor = project.property("com.mysql.cj.build.driver.version.subminor")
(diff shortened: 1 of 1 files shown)


ColinKYuen:
How does the Subminor version get updated for the next snapshot? eg if you create two snapshots in a row without manually updating `com.mysql.cj.build.driver.version.subminor`. 

It looks like what this is doing is just getting the current then plus one.
So if we have two snapshots with the same project property subversion.

sergiyvamz:
> How does the Subminor version get updated for the next snapshot? eg if you create two snapshots in a row without manually updating `com.mysql.cj.build.driver.version.subminor`.
> 
> It looks like what this is doing is just getting the current then plus one. So if we have two snapshots with the same project property subversion.

Snapshots are mutable by definition (comparing to releases that are immutable and never change after they uploaded to repository) so making another build will update the latest snapshot build. Usually snapshot builds get a next version number (minor portion of a version or a patch number) with a "-SNAPSHOT" suffix.

ColinKYuen (approved):

PR #215, opened 2022-05-26
Implement version 1 of the ReadWriteSplittingPlugin

aaron-congo:
### Summary

Implement version 1 of the ReadWriteSplittingPlugin

### Description

- connection switches to writer/reader instance on calls to setReadOnly
- requires desired instances to be defined in the connection string. The first instance in the list must be the writer instance
- currently is not compatible with the FailoverConnectionPlugin or NodeMonitoringConnectionPlugin
- limitation: ConnectionProxy#setCurrentConnection closes the old connection, so a new connection has to be opened whenever setReadOnly is called with a new value

### Additional Reviewers

@sergiyv-bitquill


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+            this.readerConnection = currentConnection;
+          }
+        }
+      }
+    } else {
+      if (!isWriterConnection() || currentConnection.isClosed()) {
+        try {
+          switchToWriterConnection(currentConnection);
+        } catch (SQLException e) {
+          throw new SQLException(
```
sergiyv-bitquill, line 86: Do we need to check if connection is open or not, similar to line 66?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+
+  @Override
+  public Object execute(Class<?> methodInvokeOn, String methodName, Callable<?> executeSqlFunc, Object[] args) throws Exception {
+    if(METHOD_SET_READ_ONLY.equals(methodName)) {
+      switchConnectionIfRequired((boolean) args[0]);
+    }
+    return this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+  }
+
+  void switchConnectionIfRequired(boolean readOnly) throws SQLException {
```
sergiyv-bitquill, line 59: Should we take into account this.inTransaction? Switching connection while been in transaction shouldn't be allowed.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+    this.connectionProvider = connectionProvider;
+    this.propertySet = propertySet;
+    this.nextPlugin = nextPlugin;
+    this.logger = logger;
+  }
+
+  @Override
+  public Object execute(Class<?> methodInvokeOn, String methodName, Callable<?> executeSqlFunc, Object[] args) throws Exception {
+    if(METHOD_SET_READ_ONLY.equals(methodName)) {
+      switchConnectionIfRequired((boolean) args[0]);
```
sergiyv-bitquill, line 54: Need to check arg array. There might be no elements at all.

src/main/core-impl/java/com/mysql/cj/conf/url/AwsMultiHostConnectionUrl.java
```diff
@@ -0,0 +1,14 @@
+package com.mysql.cj.conf.url;
```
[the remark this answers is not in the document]
davecramer: Perhaps we should add a test that checks style ? One of the checks is for a license header

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+            MysqlErrorNumbers.SQL_STATE_UNABLE_TO_CONNECT_TO_DATASOURCE);
+      }
+
+      // stick with the current connection
+      if (this.readerConnection != currentConnection) {
+        if (this.readerConnection != null) {
+          this.readerConnection.close();
+        }
+        this.readerConnection = currentConnection;
+      }
```
karenc-bq, line 238: Can we create a method for the duplicated logic?

src/test/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPluginTest.java
```diff
@@ -0,0 +1,636 @@
...
+import org.mockito.Mockito;
+
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Properties;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.AdditionalMatchers.not;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
```
karenc-bq, line 21: Can we avoid import *?

src/test/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPluginTest.java
```diff
@@ -0,0 +1,636 @@
...
+  private static final int WRITER_INDEX = 0;
+
+  @Test
+  public void testHostInfoStored() throws SQLException {
+    String url = "jdbc:mysql:aws://writer,reader1,reader2/test?" +
+        "connectionPluginFactories=com.mysql.cj.jdbc.ha.plugins.ReadWriteSplittingPluginFactory";
+    ConnectionUrl connUrl = ConnectionUrl.getConnectionUrlInstance(url, new Properties());
+
+    ICurrentConnectionProvider mockCurrentConnectionProvider = Mockito.mock(ICurrentConnectionProvider.class);
+    PropertySet mockPropertySet = Mockito.mock(PropertySet.class);
```
karenc-bq, line 33: https://github.com/awslabs/aws-mysql-jdbc/tree/main/src/test/java/com/mysql/cj/jdbc/ha/plugins
Is it possible to use a similar mocking style to the other plugin tests?

karenc-bq: https://frontbackend.com/java/difference-between-mockito-mock-method-and-the-mock-annotation

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+                MysqlErrorNumbers.SQL_STATE_UNABLE_TO_CONNECT_TO_DATASOURCE, e);
+          }
+
+          // stick with the current connection
+          if (this.readerConnection != currentConnection) {
+            if (this.readerConnection != null) {
+              this.readerConnection.close();
+            }
+            this.readerConnection = currentConnection;
+          }
```
karenc-bq, line 78:
```suggestion
      if (this.readerConnection != currentConnection && this.readerConnection != null) {
        this.readerConnection.close();
      }
      this.readerConnection = currentConnection;
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
aaron-congo: Good point, thanks

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+  @Override
+  public Object execute(Class<?> methodInvokeOn, String methodName, Callable<?> executeSqlFunc, Object[] args) throws Exception {
+    if(METHOD_SET_READ_ONLY.equals(methodName)) {
+      switchConnectionIfRequired((boolean) args[0]);
+    }
+    return this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+  }
+
+  void switchConnectionIfRequired(boolean readOnly) throws SQLException {
+    JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
```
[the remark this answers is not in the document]
aaron-congo: Yes, thanks for pointing that out

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+    if (readOnly) {
+      if (!isReaderConnection() || currentConnection.isClosed()) {
+        try {
+          switchToReaderConnection(currentConnection);
+        } catch (SQLException e) {
+          if(currentConnection == null || currentConnection.isClosed()) {
```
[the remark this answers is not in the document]
aaron-congo: Yep, good suggestion

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
aaron-congo: Yep, good suggestion

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
aaron-congo: I don't think so. In the line 66 scenario we have attempted to switch to a reader connection but failed, so I perform the check to see if we can just use the current connection as a fallback. This should be fine since read operations can be sent to reader or writer instances. However, in the scenario on this line, we have attempted to connect to the writer but failed. We don't know if the current connection is to a reader or writer, so we cannot use it and have to throw the exception. Let me know if I'm missing something though

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,271 @@
...
+  @Override
+  public void openInitialConnection(ConnectionUrl connectionUrl) throws SQLException {
+    this.hosts.addAll(connectionUrl.getHostsList());
+    this.nextPlugin.openInitialConnection(connectionUrl);
+    this.writerConnection = this.currentConnectionProvider.getCurrentConnection();
+  }
+
+  @Override
+  public void releaseResources() {
+    this.nextPlugin.releaseResources();
```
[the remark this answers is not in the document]
aaron-congo: good point... shouldn't we close both connections in this case or no? I don't see releaseResources being called in the ConnectionPluginManager so I'm not sure under which scenarios this would be called

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,308 @@
...
+                Messages.getString("ReadWriteSplittingPlugin.1"),
+                MysqlErrorNumbers.SQL_STATE_UNABLE_TO_CONNECT_TO_DATASOURCE, e);
+          }
+
+          // Failed to switch to a reader; use current connection as a fallback
+          changeReaderConnectionTo(currentConnection);
+        }
+      }
+    } else {
+      if (!isWriterConnection() || currentConnection.isClosed()) {
```
sergiyv-bitquill, line 111: Need to check transaction context here as well.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
sergiyv-bitquill: - You can call a method by calling it directly (like objA.setReadOnly(val)) or call it through reflection. Direct call is "controlled" by compiler and there's no way a developer passes no arguments. However reflection can be less restrictive.
- I'd keep scope of this PR for the plugin. We might fix those classes in a separate PR.
- It's a good to verify but I think Boolean would be more suitable in this case. Parameter type is array of objects so technically Boolean is that what we should expect in there.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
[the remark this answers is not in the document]
sergiyv-bitquill: The question is more about ownership of an additional connection. Current connection is owned by connection proxy while additional connections, like a reader connection in this plugin, is owned by plugin. Thus, it's a responsibility of the plugin to close and dispose all additional resources.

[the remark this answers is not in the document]
aaron-congo: Right, that makes sense. At the time releaseResources is called the plugin could have set the current connection to be either the reader or the writer, so I'll have it check and close any connections that aren't the current connection

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/IConnectionPlugin.java
```diff
@@ -28,21 +28,32 @@
...
 /**
  * Interface for connection plugins. This class implements ways to execute a JDBC method
  * and to clean up resources used before closing the plugin.
  */
 public interface IConnectionPlugin extends ITransactionContextHandler {
-  Object execute(
+  /**
+   * Execute a JDBC method on a Connection-bound Object such as a Statement or ResultSet.
+   */
+  Object executeOnConnectionBoundObject(
```
sergiyvamz, line 44: What's the issue you're trying to solve by making such major (breaking change) change in a public interface?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,328 @@
...
+  @Override
+  public Object executeOnConnection(Method method, List<Object> args) throws Exception {
+    if (METHOD_SET_READ_ONLY.equals(method.getName()) && args != null && args.size() > 0) {
+      switchConnectionIfRequired((Boolean) args.get(0));
+    }
+    return this.nextPlugin.executeOnConnection(method, args);
+  }
+
+  void switchConnectionIfRequired(Boolean readOnly) throws SQLException {
+    if (readOnly == null) return;
```
karenc-bq, line 100: I propose we follow google's checkstyle
```java
    if (readOnly == null) {
      return;
    }
```
https://checkstyle.sourceforge.io/config_blocks.html#NeedBraces

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,328 @@
...
+                MysqlErrorNumbers.SQL_STATE_UNABLE_TO_CONNECT_TO_DATASOURCE, e);
+          }
+
+          // Failed to switch to a reader; use current connection as a fallback
+          changeReaderConnectionTo(currentConnection);
+        }
+      }
+    } else {
+      if (!isWriterConnection() || currentConnection.isClosed()) {
+        if(this.inTransaction) {
```
karenc-bq, line 119:
```suggestion
        if (this.inTransaction) {
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -252,6 +253,38 @@ public Object execute(
     return result;
   }
 
+  @Override
+  public Object executeOnConnection(Method method, List<Object> args)
```
karenc-bq, line 257: Does args need to be a list? `executeOnConnectionBoundObject` uses an array. If `executeOnConnection` takes in an array instead of list we can avoid the checks and conversion later in the code
```java
    Object[] argsArray = args == null ? null : args.toArray();
    performSpecialMethodHandlingIfRequired(argsArray, methodName);
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -0,0 +1,328 @@
...
+    if (METHOD_SET_READ_ONLY.equals(method.getName()) && args != null && args.size() > 0) {
+      switchConnectionIfRequired((Boolean) args.get(0));
+    }
+    return this.nextPlugin.executeOnConnection(method, args);
+  }
+
+  void switchConnectionIfRequired(Boolean readOnly) throws SQLException {
+    if (readOnly == null) return;
+    final JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
+    if (readOnly) {
```
karenc-bq, line 102: `(!isReaderConnection() || currentConnection.isClosed())` is checked here and in `switchToReaderConnection`. Similarly for `(!isWriterConnection() || currentConnection.isClosed())` and `switchToWriterConnection`.
Is it possible to simply the conditions here to:
```java
  if (readOnly) {
    if (!this.inTransaction) {
      ...
    }
  } else {
    if (this.inTransaction) {
      ...
    }

    ...
  }
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
karenc-bq: I am a little bit confused by this check.
If `!isWriterConnection()` returned true, so we are currently on a reader connection, is it possible for `inTransaction` to be true?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
aaron-congo: The first one could work I think. If we changed it we would be entering switchToReaderConnection even when the current connection is the reader connection and it is open, in which case we don't need to switch connections, however switchToReaderConnection will detect this and return without doing anything. I don't think the second one would work though. If we changed the second check, it means that if a user was in a transaction on an open writer connection and called setReadOnly(false), even though we don't need to do anything, the error inside the inTransaction check would be thrown

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
aaron-congo: I believe you can start transactions on reader connections, for example "START TRANSACTION READ ONLY" is valid in MySQL (see [here](https://dev.mysql.com/doc/refman/5.6/en/commit.html))

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/IConnectionPlugin.java
aaron-congo: My mistake, I thought that the original execute function did not allow for the connection to be switched out by the plugins, but after testing I think I was wrong. I'll fix this today

src/demo/java/customplugins/ExecutionTimeConnectionPlugin.java
```diff
@@ -61,19 +58,17 @@ public ExecutionTimeConnectionPlugin(
     initializeTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
aaron-congo: Sorry for the slow response Dave, I forgot to reply, but thanks for the pointer! I'll create a ticket to address this

aaron-congo:
We have decided to work off of a separate feature branch so I have closed this PR and reopened a new one here: https://github.com/awslabs/aws-mysql-jdbc/pull/223

PR #218, opened 2022-05-30
chore: add workflow for stale issues

karenc-bq:
### Summary

chore: add workflow for stale issues

### Description

Add a Github action that runs every week to mark or close stale issues.

### Additional Reviewers

@hsuamz


.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
+name: Mark and Close Stale Issues
+
+on:
+  schedule:
+    # Every Monday at 1am
```
hsuamz, line 5: nit: Do we know the workflow timezone? If so, could we specify that in the comment above?

karenc-bq: addressed

hsuamz (approved):

PR #221, opened 2022-06-03
Refactor RDS connection string analysis logic into its own class

aaron-congo:
### Summary

Refactor RDS connection string analysis logic into its own class

### Description

- FailoverConnectionPlugin has connection string analysis logic that could be
required by other plugins. This commit refactors this logic into its own class
so that it can be shared
- Logic in the new class includes parsing logic to classify the RDS type of a
URL as well as logic to create cluster IDs and cluster instance templates from
the URL

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
@@ -44,13 +44,13 @@ import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.IP_ADDRESS;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.OTHER;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.RDS_CUSTOM_CLUSTER;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.RDS_INSTANCE;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.RDS_PROXY;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.RDS_READER_CLUSTER;
-import static com.mysql.cj.jdbc.ha.plugins.RdsUrl.RDS_WRITER_CLUSTER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.IP_ADDRESS;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.OTHER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_CUSTOM_CLUSTER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_INSTANCE;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_PROXY;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_READER_CLUSTER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_WRITER_CLUSTER;
 
 public class ConnectionStringTopologyProvider {
   private final Pattern auroraDnsPattern =
@@ -89,11 +89,11 @@ public class ConnectionStringTopologyProvider {
     this.initialConnectionProps = connectionProps;
   }
 
-  public RdsUrl getUrlType(ConnectionUrl connectionUrl) {
+  public RdsUrlType getUrlType(ConnectionUrl connectionUrl) {
     return getUrlType(connectionUrl.getMainHost().getHost());
   }
 
-  public RdsUrl getUrlType(String url) {
+  public RdsUrlType getUrlType(String url) {
     if (IpAddressUtils.isIPv4(url) || IpAddressUtils.isIPv6(url)) {
       return IP_ADDRESS;
     } else if (isWriterClusterDns(url)) {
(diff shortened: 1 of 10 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsUrlType.java
```diff
@@ -0,0 +1,53 @@
...
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+package com.mysql.cj.jdbc.ha.plugins;
+
+public enum RdsUrlType {
```
aaron-congo, line 29: Let me know if you think there's a better way to represent these properties

src/test/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPluginTest.java
```diff
@@ -520,14 +512,13 @@ public void testRdsProxy() throws SQLException {
   public void testRdsReaderCluster() throws SQLException {
     final ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
     final String url =
-        "jdbc:mysql:aws://my-cluster-name.cluster-XYZ.us-east-2.rds.amazonaws.com";
+        "jdbc:mysql:aws://my-cluster-name.cluster-ro-XYZ.us-east-2.rds.amazonaws.com";
```
aaron-congo, line 515: Seemed like either the name of the test or the url had to change for this test; the original test name indicated a reader cluster but the original url indicated a writer cluster. I opted for changing the url but let me know if I should instead change the name

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
```diff
@@ -0,0 +1,283 @@
...
+
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.IP_ADDRESS;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.OTHER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_CUSTOM_CLUSTER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_INSTANCE;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_PROXY;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_READER_CLUSTER;
+import static com.mysql.cj.jdbc.ha.plugins.RdsUrlType.RDS_WRITER_CLUSTER;
+
+public class ConnectionStringTopologyProvider {
```
sergiyv-bitquill, line 55: The class doesn't provide topology rather than a set of different flags and checks. I'd call it a utility class or a helper class.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
```diff
@@ -0,0 +1,283 @@
...
+  private final Pattern auroraCustomClusterPattern =
+      Pattern.compile(
+          "(.+)\\.(cluster-custom-[a-zA-Z0-9]+\\.[a-zA-Z0-9\\-]+\\.rds\\.amazonaws\\.com)",
+          Pattern.CASE_INSENSITIVE);
+  private final Pattern auroraProxyDnsPattern =
+      Pattern.compile(
+          "(.+)\\.(proxy-[a-zA-Z0-9]+\\.[a-zA-Z0-9\\-]+\\.rds\\.amazonaws\\.com)",
+          Pattern.CASE_INSENSITIVE);
+  private final Log logger;
+  private final int failoverConnectTimeoutMs;
```
sergiyv-bitquill, line 73: Do we use them?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
```diff
@@ -0,0 +1,283 @@
...
+  private final int failoverSocketTimeoutMs;
+  private final String clusterIdSetting;
+  private Map<String, String> initialConnectionProps = null;
```
sergiyv-bitquill, line 76: It seems it's the only member variable that makes this utility class stateful. Is it possible to pass it as function parameter?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
aaron-congo: The logger we use in this class to throw an error if the host pattern setting is invalid. I can convert the failoverConnectTimeout and socketTimeout to function parameters, they are stored in the cluster instance template. However, I'm not sure if its necessary to store these values in the cluster instance template or if they are used anywhere, which is maybe what your question was getting at

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ConnectionStringTopologyProvider.java
aaron-congo: Good point, yeah I'll change that

aaron-congo:
We have decided to work off of a separate feature branch for read-write splitting so I have closed this PR and reopened an equivalent one here: https://github.com/awslabs/aws-mysql-jdbc/pull/224

PR #224, opened 2022-06-06
Refactor RDS connection string analysis logic into its own class

aaron-congo:
### Summary

Refactor RDS connection string analysis logic into its own class

### Description

- FailoverConnectionPlugin has connection string analysis logic that could be
required by other plugins. This commit refactors this logic into its own class
so that it can be shared
- Logic in the new class includes parsing logic to classify the RDS type of a
URL as well as logic to create cluster IDs and cluster instance templates from
the URL

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
@@ -35,9 +35,11 @@ import com.mysql.cj.conf.PropertySet;
 import com.mysql.cj.log.Log;
 import com.mysql.cj.util.IpAddressUtils;
 import com.mysql.cj.util.StringUtils;
+import com.mysql.cj.util.Util;
 
 import java.sql.SQLException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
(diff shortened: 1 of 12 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
```diff
@@ -0,0 +1,304 @@
...
+    final int hostPort = hostInfo.getPort();
+    final RdsUrlType rdsUrlType = getUrlType(hostName);
+    final String clusterId = getClusterId(propertySet, rdsUrlType, hostName, hostPort);
+
+    String clusterInstanceTemplateHost = hostName;
+    if (rdsUrlType.isRds()) {
+      clusterInstanceTemplateHost = getRdsInstanceHostPattern(hostName);
+      if (clusterInstanceTemplateHost == null) {
+        this.logger.logError(Messages.getString("ConnectionStringTopologyProvider.5"));
+        throw new SQLException(Messages.getString("ConnectionStringTopologyProvider.5"));
```
sergiyvamz, line 114: ConnectionStringTopologyProvider - probably need to change message id

sergiyvamz (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
```diff
@@ -0,0 +1,304 @@
...
+      // "Invalid value for the 'clusterInstanceHostPattern' configuration setting - the host pattern must contain a '?'
+      // character as a placeholder for the DB instance identifiers of the instances in the cluster"
+      this.logger.logError(Messages.getString("ConnectionStringTopologyProvider.2"));
+      throw new SQLException(Messages.getString("ConnectionStringTopologyProvider.2"));
+    }
+
+    if (RDS_PROXY.equals(rdsUrlType)) {
+      // "An RDS Proxy url can't be used as the 'clusterInstanceHostPattern' configuration setting."
+      this.logger.logError(Messages.getString("ConnectionStringTopologyProvider.3"));
+      throw new SQLException(Messages.getString("ConnectionStringTopologyProvider.3"));
```
karenc-bq, line 152: I propose we store `Messages.getString` to a variable, and create a method for this log and throw logic, since it is used a few times

karenc-bq (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
aaron-congo: ah yeah my bad, thanks for catching that

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
aaron-congo: Good idea, sounds good!

PR #228, opened 2022-06-08
Add topology detection to the ReadWriteSplittingPlugin

congoamz:
### Summary

Add topology detection to the ReadWriteSplittingPlugin

### Description

- if a single host is supplied in the connection URL, attempt to query
Aurora to establish topology information
- if multiple hosts are supplied, simply use the supplied hosts for
topology information; the plugin will not query for topology

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
@@ -226,8 +226,15 @@ public class RdsHostUtils {
     return explicitlySetProperties;
   }
 
+  public int getHostIndex(List<HostInfo> hostList, HostInfo host) {
+    if (host == null) {
+      return NO_CONNECTION_INDEX;
+    }
+    return getHostIndex(hostList, host.getHostPortPair());
+  }
+
   public int getHostIndex(List<HostInfo> hostList, String hostPortPair) {
-    if (Util.isNullOrEmpty(hostList)) {
+    if (Util.isNullOrEmpty(hostList) || StringUtils.isNullOrEmpty(hostPortPair)) {
       return NO_CONNECTION_INDEX;
     }
 
@@ -240,8 +247,19 @@ public class RdsHostUtils {
     return NO_CONNECTION_INDEX;
   }
 
-  public HostInfo getHostInfo(List<HostInfo> hosts, String hostPortPair) {
-    for (HostInfo host : hosts) {
+  public HostInfo getHostInfo(List<HostInfo> hostList, HostInfo host) {
+    if (host == null) {
+      return null;
+    }
+    return getHostInfo(hostList, host.getHostPortPair());
+  }
+
+  public HostInfo getHostInfo(List<HostInfo> hostList, String hostPortPair) {
+    if (Util.isNullOrEmpty(hostList) || StringUtils.isNullOrEmpty(hostPortPair)) {
+      return null;
+    }
+
+    for (HostInfo host : hostList) {
       if (host.getHostPortPair().equals(hostPortPair)) {
         return host;
       }
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
@@ -227,7 +227,7 @@ public class RdsHostUtils {
   }
 
   public int getHostIndex(List<HostInfo> hostList, HostInfo host) {
-    if (host == null) {
+    if (Util.isNullOrEmpty(hostList) || host == null) {
       return NO_CONNECTION_INDEX;
     }
     return getHostIndex(hostList, host.getHostPortPair());
@@ -248,7 +248,7 @@ public class RdsHostUtils {
   }
 
   public HostInfo getHostInfo(List<HostInfo> hostList, HostInfo host) {
-    if (host == null) {
+    if (Util.isNullOrEmpty(hostList) || host == null) {
       return null;
     }
     return getHostInfo(hostList, host.getHostPortPair());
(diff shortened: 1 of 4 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
```diff
@@ -227,19 +227,46 @@ public Map<String, String> getExplicitlySetProperties(final PropertySet property
   }
 
   public int getHostIndex(List<HostInfo> hostList, HostInfo host) {
-    if (host == null || Util.isNullOrEmpty(hostList)) {
+    if (host == null) {
```
brunos-bq, line 230: NIT: you can leave the hostList check here, that returns immediately (even though it will return in the next stack call)

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
```diff
@@ -227,19 +227,46 @@ public Map<String, String> getExplicitlySetProperties(final PropertySet property
...
+    }
+    return getHostInfo(hostList, host.getHostPortPair());
+  }
+
+  public HostInfo getHostInfo(List<HostInfo> hostList, String hostPortPair) {
+    if (Util.isNullOrEmpty(hostList) || StringUtils.isNullOrEmpty(hostPortPair)) {
+      return null;
+    }
+
+    for (HostInfo host : hostList) {
```
brunos-bq, line 262: NIT: potentialMatch would be a good name here

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -66,24 +76,40 @@ public ReadWriteSplittingPlugin(
...
     this.connectionProvider = connectionProvider;
+    this.rdsHostUtils = rdsHostUtils;
     this.propertySet = propertySet;
     this.nextPlugin = nextPlugin;
     this.logger = logger;
   }
 
   @Override
-  public Object execute(Class<?> methodInvokeOn, String methodName, Callable<?> executeSqlFunc, Object[] args) throws Exception {
+  public Object execute(
```
brunos-bq, line 107: NIT: was this line really too long?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -141,9 +167,62 @@ private void changeReaderConnectionTo(JdbcConnection newConnection) throws SQLEx
...
+    if (currentHost == null) {
+      return;
+    }
+
+    final int currentHostIndex = this.rdsHostUtils.getHostIndex(this.hosts, currentHost.getHostPortPair());
+    if (currentHostIndex == 0) {
+      this.writerConnection = currentConnection;
+    } else if (currentHostIndex != -1) {
+      this.readerConnection = currentConnection;
+    }
```
brunos-bq, line 225: what happens if it's -1 ? should it ignore it or throw something?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -229,7 +308,8 @@ void syncSessionStateOnReadWriteSplit(JdbcConnection source, JdbcConnection targ
         return;
       }
 
-      final RuntimeProperty<Boolean> sourceUseLocalSessionState = source.getPropertySet().getBooleanProperty(PropertyKey.useLocalSessionState);
+      final RuntimeProperty<Boolean> sourceUseLocalSessionState =
```
brunos-bq, line 311: good break.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -141,9 +167,62 @@ private void changeReaderConnectionTo(JdbcConnection newConnection) throws SQLEx
...
+
+  private void updateInternalConnections(JdbcConnection currentConnection, HostInfo currentHost) {
+    if (currentHost == null) {
+      return;
+    }
+
+    final int currentHostIndex = this.rdsHostUtils.getHostIndex(this.hosts, currentHost.getHostPortPair());
+    if (currentHostIndex == 0) {
+      this.writerConnection = currentConnection;
+    } else if (currentHostIndex != -1) {
```
karenc-bq, line 223: can we use NO_CONNECTION_INDEX here

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/RdsHostUtils.java
congoamz: Yeah good point, probably makes more sense to leave it in

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: Yes, good point, thanks

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: Good question. If its -1, that means the current connection does not match any of the hosts we have information about. I'm not sure if we would ever actually hit this case. If we do, my thoughts were that the current connection is probably still usable, so we can let the user use it for now. When the user calls setReadOnly, we will switch to a host that we actually have the information for. Let me know if you think an exception would be more suitable though

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: Probably could go either way. It's broken up like this in FailoverConnectionPlugin so I just did the same here

sergiyv-bitquill (approved):

sergiyvamz (approved):

PR #246, opened 2022-07-07
pull secret if it is null before attempting to make a connection

davecramer:


karenc-bq (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
```diff
@@ -135,29 +135,24 @@ public void openInitialConnection(ConnectionUrl connectionUrl) throws SQLExcepti
...
-      attemptConnectionWithSecrets(properties, secret, connectionUrl);
-
-    } catch (SQLException connectionFailedException) {
-      // Rethrow the exception unless it was because user access was denied.
-      // In that case, retry with new credentials.
-      if (!SQLSTATE_ACCESS_ERROR.equals(connectionFailedException.getSQLState())) {
-        throw connectionFailedException;
-
+      if (secret != null) {
+        attemptConnectionWithSecrets(properties, secret, connectionUrl);
```
sergiyvamz, line 140: An original idea was to catch unsuccessful logins and refresh credentials form AWS vault. With proposed changes, the plugin no longer catches unsuccessful logins with cached credentials and doesn't trigger refresh of credentials. Thus, outdated credentials stays in the cache forever. Since cache is shared between connections, further attempts to create another connection will fail.

davecramer:
closing in favour of #248

PR #252, opened 2022-07-20
Add configuration option to load balance reader transactions

congoamz:
### Summary

Add configuration option to load balance reader transactions

### Description

- If reader load balancing is enabled and setReadOnly(true) has been called, the ReadWriteSplittingPlugin will switch reader connections at transaction boundaries
- Load balancing is performed by randomly selecting new reader hosts
- Also added a connection cache in the ReadWriteSplittingPlugin to avoid performance penalty for regular reader switching

### Additional Reviewers

@karenc-bq 
@sergiyv-bitquill 
@sergiyvamz

### DIFF
--- a/src/main/core-api/java/com/mysql/cj/conf/PropertyDefinitions.java
+++ b/src/main/core-api/java/com/mysql/cj/conf/PropertyDefinitions.java
@@ -645,7 +645,7 @@ public class PropertyDefinitions {
                 new StringPropertyDefinition(PropertyKey.serverAffinityOrder, DEFAULT_VALUE_NULL_STRING, RUNTIME_NOT_MODIFIABLE,
                         Messages.getString("ConnectionProperties.serverAffinityOrder"), "8.0.8", CATEGORY_HA, Integer.MIN_VALUE),
 
-                // Cluster-aware failover settings
+                // Connection plugin settings
 
                 new StringPropertyDefinition(PropertyKey.clusterId, DEFAULT_VALUE_NULL_STRING, RUNTIME_NOT_MODIFIABLE,
                         Messages.getString("ConnectionProperties.clusterId"), "0.1.0", CATEGORY_HA, Integer.MIN_VALUE),
@@ -712,6 +712,9 @@ public class PropertyDefinitions {
 
                 new BooleanPropertyDefinition(PropertyKey.useAwsIam, DEFAULT_VALUE_FALSE, RUNTIME_NOT_MODIFIABLE,
                         Messages.getString("ConnectionProperties.useAwsIam"), "0.3.0", CATEGORY_SECURITY, Integer.MAX_VALUE),
+
+                new BooleanPropertyDefinition(PropertyKey.loadBalanceReadOnlyTraffic, DEFAULT_VALUE_FALSE, RUNTIME_NOT_MODIFIABLE,
+                        Messages.getString("ConnectionProperties.loadBalanceReadOnlyTraffic"), "1.2.0", CATEGORY_HA, Integer.MAX_VALUE),
 
                 //
                 // CATEGORY_PERFORMANCE
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
@@ -151,7 +151,7 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
     }
 
     final boolean isAutoCommit = currentConnection.getAutoCommit();
-    if (isAutoCommit && "execute".equals(methodName) || "executeQuery".equals(methodName)) {
+    if (isAutoCommit && ("execute".equals(methodName) || "executeQuery".equals(methodName))) {
       return true;
     }
     return "commit".equals(methodName) || "rollback".equals(methodName);
@@ -459,6 +459,7 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
 
     if (!isConnectionUsable(this.readerConnection)) {
       initializeReaderConnection(currentConnection);
+      // The current connection may have changed; update it here in case it did.
       currentConnection = this.currentConnectionProvider.getCurrentConnection();
     }
 
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
@@ -156,7 +156,7 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
   }
 
   private boolean didMethodStartTransaction(String methodName, Object[] args) throws SQLException {
-    // If autocommit is off, any execute will start a transaction, unless the execute is itself closing the transaction
+    // If autocommit is off, any execute will start a transaction, unless the execute is closing the transaction using a SQL statement
     if (!this.currentConnectionProvider.getCurrentConnection().getAutoCommit()) {
       return isExecuteMethod(methodName) && !didMethodCloseTransactionUsingSqlStatement(methodName, args);
     }
@@ -183,9 +183,9 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
   }
 
   private boolean didMethodCloseTransaction(String methodName, Object[] args, boolean wasInTransactionBeforeExecution) throws SQLException {
-    // If autocommit is on, in most cases, any execute will start a transaction. The exceptions are:
-    // - We earlier started an implicit transaction via "START TRANSACTION" or "BEGIN"
-    // - The execute is itself starting a transaction via "START TRANSACTION" or "BEGIN"
+    // If autocommit is on, every execute will open and then close a transaction, except for the following cases:
+    // - We earlier started an implicit transaction using a SQL statement such as "START TRANSACTION" or "BEGIN"
+    // - The execute is starting a transaction using a SQL statement such as "START TRANSACTION" or "BEGIN"
     if (this.currentConnectionProvider.getCurrentConnection().getAutoCommit()
             && isExecuteMethod(methodName)
             && !wasInTransactionBeforeExecution
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
@@ -127,7 +127,7 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
     try {
       final Object result = this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
 
-      if (this.explicitlyReadOnly && this.loadBalanceReadOnlyTraffic && isTransactionBoundary(methodName)) {
+      if (this.explicitlyReadOnly && this.loadBalanceReadOnlyTraffic && isTransactionBoundary(methodName, args)) {
         this.inTransaction = false;
         pickNewReaderConnection();
       }
@@ -144,7 +144,7 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
     }
   }
 
-  private boolean isTransactionBoundary(String methodName) throws SQLException {
+  private boolean isTransactionBoundary(String methodName, Object[] args) throws SQLException {
     JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
     if (currentConnection == null) {
       return false;
@@ -154,6 +154,14 @@ public class ReadWriteSplittingPlugin implements IConnectionPlugin {
     if (isAutoCommit && ("execute".equals(methodName) || "executeQuery".equals(methodName))) {
       return true;
     }
+
+    if ("execute".equals(methodName) && args != null && args.length > 0) {
+        String sql = (String) args[0];
+        if (sql != null && (sql.equalsIgnoreCase("commit") || sql.equalsIgnoreCase("rollback"))) {
+          return true;
+      }
+    }
+
     return "commit".equals(methodName) || "rollback".equals(methodName);
   }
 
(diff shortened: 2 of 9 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -115,27 +123,46 @@ public Object execute(
...
   }
 
+  private boolean isTransactionBoundary(String methodName) throws SQLException {
+    JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
+    if (currentConnection == null) {
+      return false;
+    }
+
+    final boolean isAutoCommit = currentConnection.getAutoCommit();
+    if (isAutoCommit && ("execute".equals(methodName) || "executeQuery".equals(methodName))) {
```
sergiyvamz, line 154: I'm not sure about a purpose of the following condition
("execute".equals(methodName) || "executeQuery".equals(methodName))

src/test/java/testsuite/integration/container/aurora/AuroraMysqlReadWriteSplittingTest.java
```diff
@@ -523,6 +523,85 @@ public void test_multiHostUrl_topologyOverridesHostList() throws SQLException {
...
+      assertTrue(isDBInstanceWriter(writerConnectionId));
+
+      conn.setReadOnly(true);
+      conn.setAutoCommit(false);
+      String previousReaderId = queryInstanceId(conn);
+      assertNotEquals(writerConnectionId, previousReaderId);
+      assertTrue(isDBInstanceReader(previousReaderId));
+
+      for (int i = 0; i < 5; i++) {
+        conn.commit();
```
sergiyvamz, line 578: Do we need a statement that opens transaction?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -115,27 +123,46 @@ public Object execute(
     if (METHOD_SET_READ_ONLY.equals(methodName) && args != null && args.length > 0) {
       switchConnectionIfRequired((Boolean) args[0]);
     }
+
     try {
-      return this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+      final Object result = this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+
+      if (this.explicitlyReadOnly && this.loadBalanceReadOnlyTraffic && isTransactionBoundary(methodName)) {
```
sergiyvamz, line 130: It's a bit unusual to change connection after execute(), not before as I'd expect.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
sergiyvamz, line 154: Should we care about "executeUpdate", "executeLargeUpdate", "executeBatch" similar way?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -115,27 +123,46 @@ public Object execute(
...
+        this.inTransaction = false;
+        pickNewReaderConnection();
+      }
+      return result;
     } catch (SQLException e) {
       if (isCommunicationException(e)) {
```
sergiyvamz, line 136: Are we catching communication exceptions (any kind of them) or just failover exceptions? What happen if failover plugin isn't loaded?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -115,27 +123,46 @@ public Object execute(
...
+        closeAllConnections(currentConnection);
         updateTopology(currentConnection, currentHost);
-        updateInternalConnections(currentConnection, currentHost);
+        updateInternalConnectionInfo(currentConnection, currentHost);
       }
       throw e;
     }
   }
 
+  private boolean isTransactionBoundary(String methodName) throws SQLException {
```
sergiyvamz, line 147: Would it be easier to use term "pinned session/connection" instead of "transaction boundary"?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -115,27 +123,46 @@ public Object execute(
...
+    JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
+    if (currentConnection == null) {
+      return false;
+    }
+
+    final boolean isAutoCommit = currentConnection.getAutoCommit();
+    if (isAutoCommit && ("execute".equals(methodName) || "executeQuery".equals(methodName))) {
+      return true;
+    }
+    return "commit".equals(methodName) || "rollback".equals(methodName);
```
sergiyvamz, line 157: Should we also handle cases when commit and rollback are SQL statements?
statement.execute("commit")
statement.execute("rollback")

sergiyvamz:
A general note. It'd be extremely hard to troubleshoot this plugin having just driver logs. I'd suggest to add more trace messages.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: If autocommit is on, we would want to switch to a new reader after every query that is executed but not necessarily on every method called on the connection. For example if getAutoCommit was called on the connection we would not want to switch. Let me know if you think I'm missing something

src/test/java/testsuite/integration/container/aurora/AuroraMysqlReadWriteSplittingTest.java
congoamz: Sure, I'll add that in just in case

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: Good point. I thought of one potentially strange scenario that might occur with that change: with autocommit on, calling setReadOnly(true) followed by execute. After calling setReadOnly, the driver would switch to a reader connection. Then, when execute is called, the driver would switch to another reader connection before executing the statement. Is it problematic that it is abandoning the first reader connection without doing anything meaningful in this scenario?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: I don't think we should add executeUpdate and executeLargeUpdate because we should only be switching connections in read-only mode and these are write operations. However you're right that we should probably add executeBatch, good catch

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: We're handling exceptions with SQLState of 08S02 (communication link changed, I believe this SQLState was created by us and is only thrown upon failover) and 08007 (transaction resolution unknown, which is thrown when we failover inside a transaction but probably is also thrown in other scenarios). I wanted to handle only failover exceptions but this seems like the closest we can get since I'm not sure of any other way to distinguish failover exceptions from other SQLExceptions. We don't have integration tests covering the scenario where the failover plugin isn't loaded, should we add tests for that? However, according to the code, if the failover plugin isn't loaded there are two scenarios:

1. The sql state is not 08007 (most cases), in which case the exception will just be thrown without doing anything
2. The sql state is 08007, in which case this block will be entered: all connections will be closed, the driver will attempt to update its topology, and updateInternalConnectionInfo will be called (which updates the current host index and potentially the writer/reader connection to point to the current connection if it is a writer/reader).

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
congoamz: As far as I can tell the mariadb and community mysql driver do not do this but yeah we could add this in, probably a good idea

congoamz:
Agreed about the logs, I have a ticket to add logging statements to the plugin which will probably be the next task I pick up

src/main/core-api/java/com/mysql/cj/conf/PropertyDefinitions.java
```diff
@@ -713,6 +713,9 @@ public enum DatabaseTerm {
                 new BooleanPropertyDefinition(PropertyKey.useAwsIam, DEFAULT_VALUE_FALSE, RUNTIME_NOT_MODIFIABLE,
                         Messages.getString("ConnectionProperties.useAwsIam"), "0.3.0", CATEGORY_SECURITY, Integer.MAX_VALUE),
 
+                new BooleanPropertyDefinition(PropertyKey.loadBalanceReadOnlyTraffic, DEFAULT_VALUE_FALSE, RUNTIME_NOT_MODIFIABLE,
+                        Messages.getString("ConnectionProperties.loadBalanceReadOnlyTraffic"), "1.2.0", CATEGORY_HA, Integer.MAX_VALUE),
```
karenc-bq, line 717: Since the Read-write splitting plugin is a whole new feature, would the new release be v2.0.0? @hsuamz

congoamz: Chatted with Matthew and he mentioned 1.2.0 is the correct version

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -114,28 +123,114 @@ public Object execute(
...
     }
+
+    this.isTransactionBoundary = false;
+    boolean wasInTransactionBeforeExecution = this.inTransaction;
+
     try {
-      return this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+      final Object result = this.nextPlugin.execute(methodInvokeOn, methodName, executeSqlFunc, args);
+
+      if (didMethodStartTransaction(methodName, args, wasInTransactionBeforeExecution)) {
```
sergiyvamz, line 136: I feel that we can omit wasInTransactionBeforeExecution.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -114,28 +123,114 @@ public Object execute(
...
+        this.inTransaction = true;
+      } else if (didMethodCloseTransaction(methodName, args, wasInTransactionBeforeExecution)) {
```
sergiyvamz, line 138: The same here

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -114,28 +123,114 @@ public Object execute(
...
-        updateInternalConnections(currentConnection, currentHost);
+      if (isFailoverException(e)) {
+        closeAllConnections();
       }
       throw e;
+    } finally {
+      // Update connection/topology info in case the connection was changed
+      final JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
+      final HostInfo currentHost = this.currentConnectionProvider.getCurrentHostInfo();
+      updateTopology(currentConnection, currentHost);
```
sergiyvamz, line 153: So we update topology unconditionally even when connection hasn't changed?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -114,28 +123,114 @@ public Object execute(
...
+
+    if ("commit".equals(methodName) || "rollback".equals(methodName)) {
+      return true;
+    }
+
+    return didMethodCloseTransactionUsingSqlStatement(methodName, args);
+  }
+
+  private boolean didMethodCloseTransactionUsingSqlStatement(String methodName, Object[] args) {
+    if ("execute".equals(methodName)
```
sergiyvamz, line 206: executeQuery?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingPlugin.java
```diff
@@ -147,22 +242,71 @@ private void updateTopology(JdbcConnection currentConnection, HostInfo currentHo
...
+    final ArrayDeque<Integer> readerIndexes = getRandomReaderIndexes();
+    while (!readerIndexes.isEmpty()) {
+      final int index = readerIndexes.poll();
+
+      try {
+        this.readerConnection = getConnectionToHost(index);
+        final JdbcConnection currentConnection = this.currentConnectionProvider.getCurrentConnection();
+        switchCurrentConnectionTo(currentConnection, this.readerConnection, index);
+        return;
+      } catch (SQLException e) {
```
sergiyvamz, line 271: Does it makes sense to continue with while-loop in case exception is communication one, and rethrow it otherwise?

src/test/java/testsuite/integration/container/aurora/AuroraMysqlIntegrationBaseTest.java
```diff
@@ -201,7 +201,7 @@ public void setUpEach() throws InterruptedException, SQLException {
     makeSureInstancesUp(instanceIDs);
   }
 
-  protected Properties initDefaultPropsNoTimeouts() {
+  protected static Properties initDefaultPropsNoTimeouts() {
```
sergiyvamz, line 204: why?

PR #257, opened 2022-08-16
Add configuration option to load balance reader transactions

congoamz:
### Summary

Add configuration option to load balance reader transactions

### Description

- If reader load balancing is enabled and setReadOnly(true) has been called, the ReadWriteSplittingPlugin will switch reader connections at transaction boundaries
- Load balancing is performed by randomly selecting new reader hosts
- Also added a connection cache in the ReadWriteSplittingPlugin to avoid performance penalty for regular reader switching

### Additional Reviewers


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReadWriteSplittingStateMachine.java
```diff
@@ -0,0 +1,79 @@
...
+        this.logger = logger;
+    }
+
+    @Override
+    public void reset() {
+        this.currentState = READ_WRITE_STATE;
+    }
+
+    @Override
+    public void getNextState(JdbcConnection currentConnection, String methodName, Object[] args) throws SQLException {
```
sergiyvamz, line 60: Why do we need current connection as a parameter? It seems it's no use anywhere.

sergiyvamz (approved):

PR #259, opened 2022-08-23
Add logging to the ReadWriteSplittingPlugin

congoamz:
### Summary

Add logging to the ReadWriteSplittingPlugin

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq


src/main/resources/com/mysql/cj/LocalizedErrorMessages.properties
```diff
@@ -1095,15 +1095,33 @@ ClusterAwareReaderFailoverHandler.3=[ClusterAwareReaderFailoverHandler] Trying t
...
+ReadWriteSplittingPlugin.4=[ReadWriteSplittingPlugin] setReadOnly(false) was called on a read-only connection inside a transaction. Please complete the transaction before calling setReadOnly(false).
+ReadWriteSplittingPlugin.5=[ReadWriteSplittingPlugin] An error occurred while syncing session state during a read/write split.
+ReadWriteSplittingPlugin.6=[ReadWriteSplittingPlugin] The 'clusterInstanceHostPattern' configuration property is required when an IP address or custom domain is used to connect to a cluster that provides topology information.
+ReadWriteSplittingPlugin.7=[ReadWriteSplittingPlugin] The driver attempted to update its internal connection information but did not find a matching host entry for the following host in its host list: ''{0}''. Please ensure you are using the clusterInstanceHostPattern if your connection string does not follow the standard RDS format.
+ReadWriteSplittingPlugin.8=[ReadWriteSplittingPlugin] Transaction boundary detected - switching to a new reader instance.
+ReadWriteSplittingPlugin.9=[ReadWriteSplittingPlugin] {0} = {1}
+ReadWriteSplittingPlugin.10=[ReadWriteSplittingPlugin] {0} = ''{1}''
```
karenc-bq, line 1107: These two logs seem to have similar usage, do we need both of them?

sergiyvamz (approved):

src/main/resources/com/mysql/cj/LocalizedErrorMessages.properties
congoamz: Not necessarily, I used the first one for boolean values and the second for string values. For example:
- First log: `loadBalanceReadOnlyTraffic = true`
- Second log: `clusterInstanceHostPattern = 'my-host.aws.com'`

I can potentially get rid of the first one if you think that would be better, just means the boolean value would be in quotes

PR #262, opened 2022-08-24
Add documentation for the read-write splitting plugin

congoamz:
### Summary

Add documentation for the read-write splitting plugin

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,10 @@ At this point, the driver will connect to the new primary DB instance and return
 
 To the application, the logical connection will not appear to have changed, but the physical connection will have been swapped to the new primary DB instance. In this way the driver provides a faster way to reconnect to a newly promoted DB instance, thus increasing the availability of the DB cluster.
 
+### Setting the socket timeout
+
+By default, connections do not have a socket timeout. However, it is highly recommended that you set a socket timeout according to your application's needs. This can be done via the `socketTimeout` connection property. Without a socket timeout, the failover plugin may take a long time to detect a failover event.  We recommend that you set the socket timeout to a value slightly higher than the maximum time that you expect your queries to run.
+
 ### Failover Parameters
 
 In addition to [the parameters that you can configure for the MySQL Connector/J driver](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html), you can pass the following parameters to the AWS JDBC Driver through the connection URL to specify additional driver behavior.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -564,6 +564,60 @@ public class AWSSecretsManagerPluginSample {
...
+### Reader load-balancing
+
+The plugin can also load-balance queries among available reader instances by enabling the `loadBalanceReadOnlyTraffic` connection parameter. This parameter is disabled by default. To enable it, set the following connection parameter:
+
+```
+final Properties properties = new Properties();
+properties.setProperty("loadBalanceReadOnlyTraffic", "true");
+```
+
+Once this parameter is enabled, load-balancing will automatically be performed for reader instances when the connection has been set to read-only mode via `JdbcConnection#setReadOnly`. Load-balancing is performed by randomly selecting new reader instances at transaction boundaries.
```
susanmdouglas-aws, line 601: These sentences were a bit tough to read; I hope this suggestion is still technically accurate:

Once this parameter is enabled, you can enable load-balancing for reader instances by setting the `JdbcConnection#setReadOnly` property to read-only mode. Load-balancing will distribute queries across randomly selected new reader instances at transaction boundaries.

README.md
```diff
@@ -564,6 +564,60 @@ public class AWSSecretsManagerPluginSample {
...
+
+### Using the Read-Write Splitting Plugin against RDS/Aurora clusters
+
+When using the read-write splitting plugin against RDS or Aurora clusters, the plugin automatically acquires the cluster topology by querying the cluster. Because of this functionality, you do not have to supply multiple instance URLs in the connection string. Instead, users should only supply the URL for the initial instance they wish to connect to.
```
susanmdouglas-aws, line 605: How about:

Instead, supply just the URL for the initial instance to which you're connecting.

README.md
```diff
@@ -564,6 +564,60 @@ public class AWSSecretsManagerPluginSample {
...
+
+### Using the Read-Write Splitting Plugin against non-RDS clusters
+
+If you would like to use the read-write splitting plugin against a cluster that is not RDS or Aurora, the plugin will not be able to automatically acquire the cluster topology. Instead, you must supply the topology information to the plugin by specifying multiple instance URLs in the connection string. Instance URLs should be supplied via a comma-separated list. The first instance in the list must be the writer instance:
```
susanmdouglas-aws, line 609: Should the example that follows include a ':aws:' component? How about:

If you are using the read-write splitting plugin against a cluster that is not hosted on RDS or Aurora, the plugin will not be able to automatically acquire the cluster topology. Instead, you must supply the topology information in the connection string as a comma-delimited list of multiple instance URLs. The first instance in the list must be the writer instance:

README.md
congoamz: Thanks Susan, yes that is still accurate, I'll make that change

README.md
congoamz: Sounds good

README.md
congoamz: Yes it should. Even if the database is not RDS, the ':aws:' component of the protocol will enable the features that AWS has added instead of using the regular community driver features (read-write splitting, failover etc - although failover is not supported for non-RDS instances).

Good suggestion, I'll make that change!

README.md
congoamz: Ended up going with slightly different wording because I thought it was more clear/accurate. Let me know if you would like me to change it:

Once this parameter is enabled, queries will be load-balanced among reader instances after calling `setReadOnly(true)` on the Connection object. Load-balancing will switch to a new randomly selected reader instance at each transaction boundary.

susanmdouglas-aws: Sounds good, thanks!

README.md
susanmdouglas-aws: TY!

README.md
susanmdouglas-aws: TY!

hsuamz (approved):

sergiyvamz (approved):

PR #268, opened 2022-08-31
Fix bug where EFM plugin was swallowing a communications exception

congoamz:
### Summary

Fix bug where EFM plugin was swallowing a communications exception

### Description

- Connections are automatically closed when communications exceptions
  occur, so the EFM plugin should throw them when encountered.
  Otherwise, the invoked method will continue down the plugin chain and
  execute against a closed connection. This can lead to a 'No
  operations allowed after connection closed' exception

### Additional Reviewers

@sergiyv-bitquill 
@karenc-bq

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
@@ -43,6 +43,7 @@ import com.mysql.cj.log.Log;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLSyntaxErrorException;
 import java.sql.Statement;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -69,7 +70,6 @@ public class NodeMonitoringConnectionPlugin implements IConnectionPlugin {
   private final Supplier<IMonitorService> monitorServiceSupplier;
   private final Set<String> nodeKeys = new HashSet<>();
   private final ICurrentConnectionProvider currentConnectionProvider;
-  private final ConnectionMethodAnalyzer connectionMethodAnalyzer;
   private JdbcConnection connection;
 
   /**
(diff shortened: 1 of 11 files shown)


karenc-bq (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
```diff
@@ -84,9 +85,10 @@ public NodeMonitoringConnectionPlugin(
       ICurrentConnectionProvider currentConnectionProvider,
       PropertySet propertySet,
       IConnectionPlugin nextPlugin,
-      Log logger) {
+      Log logger) throws SQLException {
```
sergiyvamz, line 88: It seems that the only source of SQLException in the constructor body is generateNodeKeys(). I'd suggest to remove that call and relocate it to later stages of plugin lifecycle, like to execute().

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
```diff
@@ -316,9 +320,13 @@ private void generateNodeKeys(Connection connection) {
         }
       }
     } catch (SQLException sqlException) {
```
sergiyvamz, line 322: Would it be easier to catch SQLSyntaxException instead of performing complex checks to detect communication excpetions? The only reason to swallow exception here is a case when DB doesn't support extended Aurora functions. Vanilla Mysql doesn't have them so exception (SQLSyntaxException) should be ignored. All other exceptions should be popped up. Such solution would (maybe) simplify the code since connectionMethodAnalyzer might not be needed.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
congoamz: Sounds good

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
congoamz: Sounds good

sergiyvamz (approved):

sergiyvamz (approved):

PR #275, opened 2022-09-15
fix: unwrap nested exceptions in AWS Secrets Manager Plugin when checking if the error is caused by invalid credentials

karenc-bq:
### Summary

Unwrap nested exceptions in AWS Secrets Manager Plugin when checking if the error is caused by invalid credentials

### Additional Reviewers

@sergiyvamz

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
@@ -33,6 +33,7 @@ package com.mysql.cj.jdbc.ha.plugins;
 
 import com.mysql.cj.conf.ConnectionUrl;
 import com.mysql.cj.conf.PropertySet;
+import com.mysql.cj.exceptions.CJException;
 import com.mysql.cj.log.Log;
 import com.mysql.cj.util.LRUCache;
 
@@ -53,6 +54,7 @@ import java.util.Properties;
 import java.util.concurrent.Callable;
 
 public class AWSSecretsManagerPlugin implements IConnectionPlugin {
+
   static final String SECRET_ID_PROPERTY = "secretsManagerSecretId";
   static final String REGION_PROPERTY = "secretsManagerRegion";
   private static final String ERROR_MISSING_DEPENDENCY_SECRETS =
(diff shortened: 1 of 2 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
```diff
@@ -179,7 +179,19 @@ public void openInitialConnection(ConnectionUrl connectionUrl) throws SQLExcepti
...
-    return SQLSTATE_ACCESS_ERROR.equals(exception.getSQLState());
+
+    Throwable throwable = exception;
+    while (throwable != null) {
+      if (throwable instanceof SQLException) {
+        if (SQLSTATE_ACCESS_ERROR.equals(((SQLException) throwable).getSQLState())) {
+          return true;
+        }
+
+        throwable = throwable.getCause();
```
sergiyvamz, line 190: I think it makes sense to move this line outside of if(). We want to iterate through all nested exceptions, not just through SQLExceptions. Like this:

while(throwable != null) {
if (throwale instanceof SQLException) {
...}
throwable = throwable.getCause();
}

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
```diff
@@ -179,7 +179,19 @@ public void openInitialConnection(ConnectionUrl connectionUrl) throws SQLExcepti
    */
   private boolean isLoginUnsuccessful(SQLException exception) {
     this.logger.logTrace("Login failed. SQLState=" + exception.getSQLState(), exception);
-    return SQLSTATE_ACCESS_ERROR.equals(exception.getSQLState());
+
+    Throwable throwable = exception;
+    while (throwable != null) {
+      if (throwable instanceof SQLException) {
```
kaddybrar08, line 185: The internal exception which is wrapped in java.sql.SQLNonTransientConnectionException is of Type :- software.aws.rds.jdbc.mysql.shading.com.mysql.cj.exceptions.CJException.
This exception extends RuntimeException and not SQLException, therefore, the internal exception will be skipped and secrets won't get rotated.

The custom codefix which I tested :- 
`private boolean isLoginUnsuccessful(SQLException exception) {
    this.logger.logTrace("Login failed. SQLState=" + exception.getSQLState(), exception);
    Throwable t = exception;
    do {
      if (t instanceof SQLException) {
        if (((SQLException) t).getSQLState().equals(SQLSTATE_ACCESS_ERROR)) {
          return true;
        }
      }
      if (t instanceof CJException) {
        if (((CJException) t).getSQLState().equals(SQLSTATE_ACCESS_ERROR)) {
          return true;
        }
      }
        t = t.getCause();
    } while (t != null);
    return false;
  }`


-- Anmolpreet Singh (Chegg Inc)

sergiyvamz (approved):

Issue #214, opened 2022-05-25
After failover sometimes connections are readonly when should not. 


tommisc:
**Describe the issue**
I am getting sometimes error The MySQL server is running with the --read-only option so it cannot execute this statement
after failover for connections that are using cluster endpoint

**Driver Version?** 
1.0.0

**Java Version?**
Jdk 17

**OS Version?**

**MySQL Version?**
5.7.mysql_aurora.2.10.2

**To Reproduce**
I am using Hikari connection pool 3.4.5.
Sometimes after failover connections using writable endpoint are getting error that connection is read only. And situation will not solve over time.

**Expected behaviour**
Connection should be always writable.

**Logs**
If possible MySQL logs surrounding the occurrence of the issue

zixubingfeng:
I am having the same problem. change Hikari connection pool 2.7.9 is ok.
https://github.com/awslabs/aws-mysql-jdbc/issues/209

tommisc:
Problem also exists with Hikari version 5.0.1

karenc-bq:
Hi all, thank you for raising this issue. We're currently looking into it and will share more information once we know more about what is going on. Thank you for your patience.

Michael-S-Smith:
Any update on this? We have a 3-minute failover duration and are unsure if this is a problem with Hikari or the driver. Does the fact the bug tag was added mean this has been found to be a driver bug?

sergiyvamz:
Hi @Michael-S-Smith 

Unfortunately there's no update at the moment. We're still looking into the issue. We're not yet sure if the root cause of the issue is a pure driver bug or an integration issue between the driver and Hikari CP. 

We'll keep you informed. Thank you for patience!

dipindev:
This seems to be an issue in 4.0.3 as well. Please confirm if there is any new releases coming up which involves the fix.

sergiyvamz:
Hello @tommisc 

We have investigated the issue and found a single scenario that may lead to the issue that you reported. If your application performs steps as it follows, then driver failover process connects to a replica instead of a writer node.

```
// open connection to a writer node by using DriverManager or HikariCP

// application uses connection

connection.setReadOnly(true);

// application uses connection

Statement statement = connection.createStatement();

// failover occurs here or during executing the following query
statement.executeQuery("select sleep(1000)");

// failover is completed and connection gets connected to a reader node

```

[...]

PR #302, opened 2022-10-27
fix: after Aurora failover, failover is disabled in FailoverConnectionPlugin and re-connection can target the wrong host

albanf:
### Summary

This PR fixes several issues during re-connection after failover:
* if the topology has only one host (this happens after a replica is promoted to writer, and before the previous writer has restarted
* `FailoverConnectionPlugin.pickNewConnection` is using stale topology, so it does not reconnect to the expected host
* if hosts are switched (replica became writer, writer became replica), `FailoverConnectionPlugin.updateHostIndex` will keep the connection to the same host, although that host is now the replica

### Description

Most changes are clear in `FailoverConnectionPlugin`, and covered by unit tests.
I've added logging in several places in order to investigate this issue, and that includes the stack trace when the same messages are logged in multiple places. debug and trace logging are enclosed in tests to check the logging level first.

### DIFF
--- a/src/main/core-api/java/com/mysql/cj/conf/HostInfo.java
+++ b/src/main/core-api/java/com/mysql/cj/conf/HostInfo.java
@@ -34,6 +34,7 @@ import static com.mysql.cj.util.StringUtils.isNullOrEmpty;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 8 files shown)


albanf:
Also fixes #214

albanf:
Hello, thanks for looking at the PR.
I'll pulled the changes and run some tests, and eventually rebase the PR on main if needed.

I think they are still some issues and race conditions with these fixes
* the changes in #306 may increase the recovery time for the writer: if the secondary node is very slow to come up, or never comes up, then it will loop longer/forever and never recover
* the `isFailoverEnabled` method relies on the list of hosts from the current topology, so if a new connection is made during failover, and the topology is refreshed from the database by this new connection (let's say topology in the cache has expired), that connection will store a unique host in the list and will have failover disabled
* same if we scale down a cluster to 1 node then scale it up, all the connections will have failover disabled and will never try to get a newer topology
* if the hosts were switched after a failover, there is still the issue in updateHostIndex that the connection is deemed valid although the position of the host has changed (i.e. its role), look for
```// found the same node at different position in the topology
// adjust current index only; connection is still valid
this.currentHostIndex = latestHostIndex;
```

Most of these issues happen in the context of a pool, where some connections might failover, and other might only be used once the failover is done, and the cached topology has changed.

I''ll let you know what I find, probably next Monday.

alecc-bq:
Hi @albanf,

Just wanted to let you know that we are currently looking into the four issues you have raised here. Thank you for your patience!

crystall-bitquill:
Hi @albanf,

Apologies for the delayed response.

For this PR, many changes have been made to the driver since this issue was raised, and rebasing now may be more difficult than opening a new PR. Therefore, I've opened a new PR (#426) with similar changes to address the issue, and you have been tagged for visibility.

Regarding your additional concerns,

1. The changes made in #306 were meant to ensure connections were made to writers instead of stale readers in the event of a failover. Although it is true that writer recovery time may be increased, the amount of time spent waiting for failover operations to complete can still be controlled with parameters like [`failoverTimeoutMs` or `failoverReaderConnectTimeoutMs`](https://github.com/awslabs/aws-mysql-jdbc#failover-parameters).

2-3. For your second and third points, some changes have been made in PR #345 to address these problems. These changes prevent connections from having failover disabled after scaling down to or starting with a single node. When the failover plugin executes statements, the topology will be updated if possible and failover will be enabled.

4. Some changes have been made in PR #426  to address the host role changes.

Thank you again for your feedback in identifying these improvements, as well as for your patience. Please feel free to make contributions again in the future!

PR #309, opened 2022-11-14
Fix #308: Statements with same ConnectionProxy should be distinct

### LINKED ISSUE #308: Distinct ClientPreparedStatement objects incorrectly have same hashCode and match with equals()
### Describe the bug

Distinct `ClientPreparedStatement`s sharing a `ConnectionProxy` will all use the `hashCode()` and `equals()` of the `ConnectionProxy` and therefore will all have the same hash code and be equals via the `equals()` call. 

This is a problem not only because fundamentally the objects are different so they should have different hash codes and not be equal, but also because when using the AWS Aurora MySQL driver with Hibernate 5, the `ClientPreparedStatement`s will be registered in `ResourceRegistryStandardImpl` here https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/resource/jdbc/internal/ResourceRegistryStandardImpl.java#L87 . This code will throw an exception when the `Statement` (key) was already registered. Multiple different statements with the same `hashCode()` that are equal via the `equals()` will cause this problem.

### Expected Behavior

Multiple `ClientPreparedStatement`s using the same `ConnectionProxy` should have different hash code values and not be equal via `equals()`

### Current Behavior

Normally, Hibernate code will throw an exception when the `Statement` (key) was already registered. It detects this by seeing if the key already exists in its map. Using the AWS Aurora MySQL driver, multiple different statements with the same `hashCode()` that are equal via the `equals()` will cause a "JDBC Statement already registered" even if the `Statement`s are not the same.

### Reproduction Steps

Using the same Hibernate `Session`, call `prepareQueryStatement()` with two different `Statement`s that use different SQL queries. 

Expected: Successful registration of the two `Statement`s

Actual: First `ClientPreparedStatement` registers successfully but the second registration has an `HibernateException` thrown from `ResourceRegistryStandardImpl` with message `JDBC Statement already registered`

### Possible Solution

One possible solution is to simply not let the proxy directly execute `equals()` and `hashCode()` and instead defer to the underlying Statement.

`src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java` implements `invoke()` (https://github.com/awslabs/aws-mysql-jdbc/blob/main/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java#L312) which will detect if the method is `equals` or `hashCode` (https://github.com/awslabs/aws-mysql-jdbc/blob/main/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java#L296) and if so, it uses the implementation of the `ConnectionProxy` instead of the implementation of the underlying object (https://github.com/awslabs/aws-mysql-jdbc/blob/main/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java#L267). In this case, the underlying object is the `ClientPreparedStatement`. Two different `ClientPreparedStatement` objects that share the same `ConnectionProxy` will be equal because they share the same hash code and `equals()` will evaluate to true.

### Additional Information/Context

_No response_

### The AWS JDBC Driver for MySQL version used

1.1.1 (and 1.1.0)

### JDK version used

1.8

### Operating System and version

MacOS Monterey 12.3.1

buitcj:
Fix hashCode and equals methods for ClientPreparedStatements created by the same ConnectionProxy. Distinct Statements that share the same ConnectionProxy should not have the same hash code value and should not be equal via the equals() method.

### Summary

See https://github.com/awslabs/aws-mysql-jdbc/issues/308

Fix hashCode and equals methods for ClientPreparedStatements created by the same ConnectionProxy. Distinct Statements that share the same ConnectionProxy should not have the same hash code value and should not be equal via the equals() method.

### Description

Removed code to use the equals and hashcode implementations of the ConnectionProxy object and instead deferring to the underlying object's implementation.

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
@@ -64,8 +64,6 @@ public class ConnectionProxy implements ICurrentConnectionProvider, InvocationHa
 
   /** Null logger shared by all connections at startup. */
   protected static final Log NULL_LOGGER = new NullLogger(Log.LOGGER_INSTANCE_NAME);
-  static final String METHOD_EQUALS = "equals";
-  private static final String METHOD_HASH_CODE = "hashCode";
   private final JdbcPropertySetImpl connProps = new JdbcPropertySetImpl();
   /** The logger we're going to use. */
   protected transient Log log = NULL_LOGGER;
@@ -183,10 +181,6 @@ public class ConnectionProxy implements ICurrentConnectionProvider, InvocationHa
       throws Throwable {
     final String methodName = method.getName();
 
-    if (isDirectExecute(methodName)) {
-      return executeMethodDirectly(methodName, args);
-    }
-
     Object[] argsCopy = args == null ?  null : Arrays.copyOf(args, args.length);
 
     try {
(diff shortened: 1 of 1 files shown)


buitcj:
Still testing

PR #314, opened 2022-11-21
Fix getBigInteger error message

buitcj:
### Summary

Error key `ResultSet.Bad_format_for_BigInteger` expects the unexpected value param first then the column index second: `Bad format for BigInteger ''{0}'' in column {1}.` (https://github.com/awslabs/aws-mysql-jdbc/blob/main/src/main/resources/com/mysql/cj/LocalizedErrorMessages.properties#L446)

However, we incorrectly log the parameters in the incorrect order.

### Description

This PR fixes the order of the params for NFEs for `getBigInteger`().


src/main/user-impl/java/com/mysql/cj/jdbc/result/ResultSetImpl.java
```diff
@@ -841,7 +841,7 @@ public BigInteger getBigInteger(int columnIndex) throws SQLException {
             return new BigInteger(stringVal);
         } catch (NumberFormatException nfe) {
             throw SQLError.createSQLException(
-                    Messages.getString("ResultSet.Bad_format_for_BigInteger", new Object[] { Integer.valueOf(columnIndex), stringVal }),
+                    Messages.getString("ResultSet.Bad_format_for_BigInteger", new Object[] { stringVal, Integer.valueOf(columnIndex) }),
```
sergiyvamz, line 844: Can we check the same for BigDecimal?

sergiyvamz (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/result/ResultSetImpl.java
jubui: We do not do a similar `getString` call for `getBigInteger` and we don't have a parameterized error for it either

src/main/resources/com/mysql/cj/LocalizedErrorMessages.properties
```diff
@@ -444,7 +444,7 @@ ReplicationGroupManager.0=Unable to register replication host management bean wi
 ResultSet.Retrieved__1=Retrieved 
 ResultSet.Bad_format_for_BigDecimal=Bad format for BigDecimal ''{0}'' in column {1}.
 ResultSet.Bad_format_for_BigInteger=Bad format for BigInteger ''{0}'' in column {1}.
-ResultSet.Column_Index_out_of_range_low=Column Index out of range, {0} < 1.
+ResultSet.Column_Index_out_of_range_low=Column Index out of range, {0} < {1}.
```
jubui, line 447: @sergiyvamz found an problem with an unparameterized part of another error code here

src/main/user-impl/java/com/mysql/cj/jdbc/result/ResultSetImpl.java
buitcj: We do not do a similar `getString` call for `getBigInteger` and we don't have a parameterized error for it either

src/main/resources/com/mysql/cj/LocalizedErrorMessages.properties
buitcj, line 447: @sergiyvamz I found an problem with an unparameterized part of another error code here

buitcj:
@sergiyvamz Anything required or suggested for me in order to get this merged?

Issue #318, opened 2022-12-06
DatabaseMetaData#getDriverVersion is reporting a SNAPSHOT version


buitcj:
### Describe the bug

Official v1.1.2 release of the driver should report a non-snapshot version via `DatabaseMetaData#getDriverVersion` but instead reports a snapshot version.

### Expected Behavior

Driver version should report a non snapshot version like `aws-mysql-connector-java-1.1.2`

### Current Behavior

Driver version reports a snapshot version like `aws-mysql-connector-java-1.1.2-SNAPSHOT (Revision: d85fd009cb21f60b0a9316b94b05628633e261f7)`

![image](https://user-images.githubusercontent.com/55841907/205889789-afbbe05d-a0c9-47cc-85df-f0b56bf5e4e1.png)

### Reproduction Steps

With a valid aurora mysql instance and credentials, 

```
Connection conn = DriverManager.getConnection(jdbcUrl, userName, password);
System.out.println(conn.getMetaData().getDriverVersion());
```

with a gradle build file using maven central and `implementation 'software.aws.rds:aws-mysql-jdbc:1.1.2'`

### Possible Solution

Possible problem with release process?

### The AWS JDBC Driver for MySQL version used

1.1.2

### JDK version used

1.8

### Operating System and version

MacOS Monterey 12.6.1

hsuamz:
Thanks for the report. We'll take a look at this.

karenc-bq:
Thank you for raising this issue, the bug has been fixed.

PR #321, opened 2022-12-06
chore: remove SNAPSHOT from the full version

karenc-bq:
### Summary

Remove the SNAPSHOT string from the versioning

### Description

`DatabaseMetaData#getDriverVersion` returns the full product name and version with "SNAPSHOT" attached, e.g. `aws-mysql-connector-java-1.1.2-SNAPSHOT` 

Remove the SNAPSHOT string from the versioning.

Addresses #318 

### Additional Reviewers

@aaron-congo 
@hsuamz

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -267,7 +267,7 @@ tasks.register<Sync>("replaceTokens") {
     val git = org.ajoberstar.grgit.Grgit.open(mapOf("currentDir" to project.rootDir))
     val revision = git.head().id
 
-    val versionFull = "$versionMajor.$versionMinor.$versionSubminor-SNAPSHOT"
+    val versionFull = "$versionMajor.$versionMinor.$versionSubminor"
 
     val fullProdName = "${project.property("com.mysql.cj.build.driver.name")}-$versionFull"
 


aaron-congo (approved):

congoamz (approved):

davecramer:
One option here is to use a CommentPreprocessorTask as per:
https://github.com/pgjdbc/pgjdbc/blob/master/build.gradle.kts#L304-L321

build.gradle.kts
```diff
@@ -267,7 +267,7 @@ tasks.register<Sync>("replaceTokens") {
     val git = org.ajoberstar.grgit.Grgit.open(mapOf("currentDir" to project.rootDir))
     val revision = git.head().id
 
-    val versionFull = "$versionMajor.$versionMinor.$versionSubminor-SNAPSHOT"
+    val versionFull = "$versionMajor.$versionMinor.$versionSubminor"
```
sergiyvamz, line 270: Should we use variable "version" from line 38?

davecramer: seems like that would be a better option.

Issue #304, opened 2022-10-27
Sources jar does not contain relocated classes


albanf:
### Describe the issue

The classes in the sources jar are not the same as the classes in the actual jar.
This is because package `com.mysql` is relocated to `software.aws.rds.jdbc.mysql.shading.com.mysql` during build, but not when sources are packaged.

### Links

N/A

karenc-bq:
Hi @albanf, the `aws-mysql-jdbc-1.1.1.jar` we released gets shaded during compilation. The `com.mysql` package itself are not relocated in the source code, so the sources JAR match what we have in the source code. Is there a specific reason to also have the sources JAR shaded as well?

albanf:
When debugging the failover issues, I downloading the maven sources to step through the code, but then the source files do not match the runtime classes, so it's difficult to set breakpoints and run a debugging session

karenc-bq:
I see, thank you for raising this, we will investigate it.

weaselmetal:
For reference, I've brought this up before. https://github.com/awslabs/aws-mysql-jdbc/issues/239

aaronchung-bitquill:
Hello @albanf 
The solution to this has been merged. Could you test out our 1.1.3-Snapshot build [here](https://aws.oss.sonatype.org/content/repositories/snapshots/software/aws/rds/aws-mysql-jdbc/) and let us know if the issue still persists? 
Thank you.

albanf:
I can confirm that this is working as expected after this commit, thanks

PR #325, opened 2022-12-16
build: add shading to -sources.jar to reflect the main jar which is shaded (#304)

aaronchung-bitquill:
### Summary

build: add shading to -sources.jar to reflect the main jar which is shaded (#304)
NOTE: marked as WIP since I've yet to finish all the testing. Creating it for early feedback.

### Description

Currently, the sources JAR is not shaded, but the main JAR is. This makes it difficult for users that are debugging using the source files in their IDEs. The purpose of this PR is to address that by shading the sources JAR as well so that they match the main JAR. 
The Gradle shadow plugin does not support shading the sources JAR. So a custom solution was instead made.

Changes:
- Added a new Gradle task `shadeSourcesJar` to shade the sources JAR
- Made the `assemble` task depend on the `shadeSourcesJar` task
- Replaced the `sourcesJar` task artifacts in the Maven publication with the `shadeSourcesJar` task artifacts

### Additional Reviewers

@karenc-bq @aaron-congo @sergiyv-improving

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,12 +30,18 @@
 */
 
 import org.apache.tools.ant.filters.ReplaceTokens
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
 
 // Driver version numbers
 val versionMajor = project.property("com.mysql.cj.build.driver.version.major")
 val versionMinor = project.property("com.mysql.cj.build.driver.version.minor")
 val versionSubminor = Integer.parseInt(project.property("com.mysql.cj.build.driver.version.subminor").toString()) + if (project.property("snapshot") == "true") 1 else 0
 version = "$versionMajor.$versionMinor.$versionSubminor" + if (project.property("snapshot") == "true") "-SNAPSHOT" else ""
+
+// Shared Shading Values
+val shadingOriginPackage: String = "com.mysql"
+val shadingDestinationPackage: String = "software.aws.rds.jdbc.mysql.shading.com.mysql"
 
 plugins {
     base
@@ -126,7 +132,7 @@ tasks.shadowJar {
         exclude(dependency(":"))
     }
 
-    relocate ("com.mysql", "software.aws.rds.jdbc.mysql.shading.com.mysql")
+    relocate (shadingOriginPackage, shadingDestinationPackage)
 
     exclude("instrumentation/**")
     exclude("demo/**")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -150,6 +156,44 @@ tasks.register<Jar>("cleanShadowJar") {
...
+
+    val sourceCharset: java.nio.charset.Charset = StandardCharsets.UTF_8
+    val shadedSourcesDir: java.nio.file.Path = file(tmpShadowedSourcesPath).toPath()
+
+    doFirst {
+        zipTree(tasks["sourcesJar"].outputs.files.asPath).visit {
+            val relativePath = this.relativePath.toString()
+
+            if (this.file.isDirectory) {
+                return@visit
```
aaron-congo, line 176: Just for my own understanding, what is @visit? I haven't seen this before

aaronchung-bitquill: In Kotlin, if we have nested functions, the `@` helps explicitly state which nested function to return from. 
Found a decent explanation here https://stackoverflow.com/questions/40160489/kotlin-whats-does-return-mean

build.gradle.kts
```diff
@@ -423,4 +468,4 @@ tasks.withType<Test> {
         this.showStandardStreams = true
     }
     useJUnitPlatform()
-}
+}
```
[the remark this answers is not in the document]
aaronchung-bitquill: Done!

build.gradle.kts
```diff
@@ -150,6 +156,44 @@ tasks.register<Jar>("cleanShadowJar") {
     }
 }
 
+tasks.register<Jar>("shadeSourcesJar") {
+//    archiveClassifier.set("shaded-sources")
```
aaron-congo, line 160: should we remove this comment?

build.gradle.kts
```diff
@@ -150,6 +156,44 @@ tasks.register<Jar>("cleanShadowJar") {
...
+    archiveClassifier.set("sources")
+    dependsOn("sourcesJar")
+
+    val shadingOriginPath: String = shadingOriginPackage.replace('.', '/')
+    val shadingDestinationPath: String = shadingDestinationPackage.replace('.', '/')
+    val tmpShadowedSourcesPath = "${buildDir}/tmp/shadowedSources"
```
sergiyvamz, line 166: Any chance we get some files in this folder left from a previous build?

sergiyvamz (approved):

build.gradle.kts
aaronchung-bitquill: Good catch! Definitely.
Yeah that is something I missed and will need to add.

build.gradle.kts
aaronchung-bitquill: Ah good catch. Removed!

build.gradle.kts
aaronchung-bitquill: Done!

PR #356, opened 2023-01-31
Optimize thread locks and expiring cache for Enhanced Monitoring plugin

sergiyvamz:
### Summary

Optimize thread locks and expiring cache for Enhanced Monitoring plugin.

### Description

- Expiring cache object has been replaced with CacheMap<>
- AuroraTopologyService uses multiple caches that provides a better performance and concurrency.
- Optimized work with monitoring context in Enhanced Monitoring plugin. A single queue with monitoring context is replaced with a two separate queues. The first queue holds all active context. When context monitoring time comes, the context is relocated to the second queue. Separating queues help to clear identify context that requires active node monitoring.
- Performance optimization for various part of the code.

### Additional Reviewers

@karenc-bq 
@congoamz

### DIFF
--- a/src/main/core-api/java/com/mysql/cj/util/CacheMap.java
+++ b/src/main/core-api/java/com/mysql/cj/util/CacheMap.java
@@ -124,13 +124,10 @@ public class CacheMap<K,V> {
       }
       CacheItem<?> other = (CacheItem<?>) obj;
       if (item == null) {
-        if (other.item != null) {
-          return false;
-        }
-      } else if (!item.equals(other.item)) {
-        return false;
+        return other.item == null;
+      } else {
+        return item.equals(other.item);
       }
-      return true;
     }
 
     @Override
--- a/src/main/core-api/java/com/mysql/cj/util/CacheMap.java
+++ b/src/main/core-api/java/com/mysql/cj/util/CacheMap.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License, version 2.0
+ * (GPLv2), as published by the Free Software Foundation, with the
+ * following additional permissions:
+ *
+ * This program is distributed with certain software that is licensed
+ * under separate terms, as designated in a particular file or component
+ * or in the license documentation. Without limiting your rights under
+ * the GPLv2, the authors of this program hereby grant you an additional
+ * permission to link the program and your derivative works with the
+ * separately licensed software that they have included with the program.
+ *
+ * Without limiting the foregoing grant of rights under the GPLv2 and
+ * additional permission as to separately licensed software, this
+ * program is also subject to the Universal FOSS Exception, version 1.0,
+ * a copy of which can be found along with its FAQ at
+ * http://oss.oracle.com/licenses/universal-foss-exception.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License, version 2.0, for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see
+ * http://www.gnu.org/licenses/gpl-2.0.html.
+ */
+
+package com.mysql.cj.util;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class CacheMap<K,V> {
+
+  private final Map<K, CacheItem<V>> cache = new ConcurrentHashMap<>();
+  private final long cleanupIntervalNanos = TimeUnit.MINUTES.toNanos(10);
+  private AtomicLong cleanupTimeNanos = new AtomicLong(System.nanoTime() + cleanupIntervalNanos);
+
+  public CacheMap() {
+  }
+
+  public V get(final K key, final long itemExpirationNano) {
+    CacheItem<V> cacheItem = cache.computeIfPresent(key,
+        (kk, vv) -> vv.isExpired()
+            ? null
+            : new CacheItem<>(vv.item, System.nanoTime() + itemExpirationNano));
+    return cacheItem == null ? null : cacheItem.item;
+  }
+
+  public V get(final K key, final V defaultItemValue, long itemExpirationNano) {
+    CacheItem<V> cacheItem = cache.compute(key,
+        (kk, vv) -> new CacheItem<>(
+            (vv == null || vv.isExpired() ? defaultItemValue : vv.item),
+            System.nanoTime() + itemExpirationNano));
+    return cacheItem.item;
+  }
+
+  public void put(final K key, final V item, long itemExpirationNano) {
+    cache.put(key, new CacheItem<>(item, System.nanoTime() + itemExpirationNano));
+    cleanUp();
+  }
+
+  public void putIfAbsent(final K key, final V item, long itemExpirationNano) {
+    cache.putIfAbsent(key, new CacheItem<>(item, System.nanoTime() + itemExpirationNano));
+    cleanUp();
+  }
+
+  public void remove(final K key) {
+    cache.remove(key);
+    cleanUp();
+  }
+
+  public void clear() {
+    cache.clear();
+  }
+
+  public int size() { return this.cache.size(); }
+
+  private void cleanUp() {
+    if (this.cleanupTimeNanos.get() < System.nanoTime()) {
+      this.cleanupTimeNanos = new AtomicLong(System.nanoTime() + cleanupIntervalNanos);
+      cache.entrySet().forEach(entry -> {
+        if (entry.getValue() == null || entry.getValue().isExpired()) {
+          cache.remove(entry.getKey());
+        }
+      });
+    }
+  }
+
+  private static class CacheItem<V> {
+    final V item;
+    final long expirationTime;
+
+    public CacheItem(V item, long expirationTime) {
+      this.item = item;
+      this.expirationTime = expirationTime;
+    }
+
+    boolean isExpired() {
+      return System.nanoTime() > expirationTime;
+    }
+
+    @Override
+    public int hashCode() {
+      final int prime = 31;
+      int result = 1;
+      result = prime * result + ((item == null) ? 0 : item.hashCode());
+      return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (this == obj) {
+        return true;
+      }
+      if (obj == null) {
+        return false;
+      }
+      if (getClass() != obj.getClass()) {
+        return false;
+      }
+      CacheItem<?> other = (CacheItem<?>) obj;
+      if (item == null) {
+        if (other.item != null) {
+          return false;
+        }
+      } else if (!item.equals(other.item)) {
+        return false;
+      }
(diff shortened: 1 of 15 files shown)


src/main/core-api/java/com/mysql/cj/util/CacheMap.java
```diff
@@ -0,0 +1,165 @@
...
+package com.mysql.cj.util;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class CacheMap<K,V> {
+
+  private final Map<K, CacheItem<V>> cache;
```
karenc-bq, line 41: Since this is final can we just initialize the map here
```java
private final Map<K, CacheItem<V>> cache =  new ConcurrentHashMap<>();
```

src/main/core-api/java/com/mysql/cj/util/CacheMap.java
```diff
@@ -0,0 +1,165 @@
...
+        return false;
+      return true;
+    }
+
+    @Override
+    public String toString() {
+      return "CacheItem [item=" + item + ", expirationTime=" + expirationTime + "]";
+    }
+  }
+}
```
karenc-bq, line 165:
```suggestion
}

```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/AuroraTopologyService.java
```diff
@@ -190,30 +180,25 @@ public void setClusterInstanceTemplate(HostInfo clusterInstanceTemplate) {
   @Override
   public List<HostInfo> getTopology(JdbcConnection conn, boolean forceUpdate)
       throws SQLException {
-    ClusterTopologyInfo clusterTopologyInfo = topologyCache.get(this.clusterId);
 
-    if (clusterTopologyInfo == null
-        || Util.isNullOrEmpty(clusterTopologyInfo.hosts)
-        || forceUpdate
-        || refreshNeeded(clusterTopologyInfo)) {
+    List<HostInfo> hosts = topologyCache.get(this.clusterId, this.refreshRateNanos);
```
congoamz, line 184: It looks like the CacheMap `get` implementation will keep on updating the expiration time of the entry each time we get here. This means if we keep fetching the cached item before the expiry, we will keep on updating the expiration time and returning the cached item. Is this what we want? I would've expected we set the expiry time once only, when the entry is first created. That way we would avoid returning a cached item that could be very old if we've been fetching it from the cache regularly

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
```diff
@@ -175,36 +177,37 @@ public void setCurrentConnection(JdbcConnection connection, HostInfo info) {
...
-          () -> method.invoke(currentConnection, args),
-          argsCopy);
-      return proxyIfReturnTypeIsJdbcInterface(method.getReturnType(), result);
-    } catch (Exception e) {
-      // Check if the captured exception must be wrapped by an unchecked exception.
-      Class<?>[] declaredExceptions = method.getExceptionTypes();
-      for (Class<?> declaredException : declaredExceptions) {
-        if (declaredException.isAssignableFrom(e.getClass())) {
-          throw e;
+    synchronized (currentConnection) {
```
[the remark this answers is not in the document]
sergiyvamz: We don't support such scenario. Each connection can be shared by threads but only one thread can actively use it. That's why we need to synchronize on current connection.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
```diff
@@ -62,23 +60,22 @@ static class ConnectionStatus {
...
   private final Log logger;
   private final PropertySet propertySet;
   private final HostInfo hostInfo;
   private final IMonitorService monitorService;
-  private final AtomicLong connectionCheckIntervalMillis = new AtomicLong(DEFAULT_CONNECTION_CHECK_INTERVAL_MILLIS);
-  private final AtomicLong contextLastUsedTimestampNano = new AtomicLong();
-  private final AtomicBoolean stopped = new AtomicBoolean(true);
-  private final AtomicBoolean isConnectionCheckIntervalInitialized = new AtomicBoolean(false);
+  private volatile long contextLastUsedTimestampNano;
+  private boolean stopped = false;
```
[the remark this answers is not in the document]
sergiyvamz: It's not necessary here. I cleaned it up.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
```diff
@@ -139,42 +127,114 @@ public void stopMonitoring(MonitorConnectionContext context) {
...
+
+              // If context is still valid and node is still healthy, it needs to continue updating this context
+              if (monitorContext.isActiveContext() && !monitorContext.isNodeUnhealthy()) {
+                this.activeContexts.add(monitorContext);
+                if (firstAddedMonitorContext == null) {
+                  firstAddedMonitorContext = monitorContext;
+                }
+
+                if (delayMillis == -1 || delayMillis < monitorContext.getFailureDetectionIntervalMillis()) {
+                  delayMillis = monitorContext.getFailureDetectionIntervalMillis();
```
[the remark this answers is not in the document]
sergiyvamz: Good point. Thanks!

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
```diff
@@ -139,42 +127,114 @@ public void stopMonitoring(MonitorConnectionContext context) {
...
+              delayMillis = MIN_CONNECTION_CHECK_TIMEOUT_MILLIS;
+            }
+            // Use this delay as node checkout timeout since it corresponds to min interval for all active contexts
+            this.nodeCheckTimeoutMillis = delayMillis;
           }
 
-          TimeUnit.MILLISECONDS.sleep(
-                  Math.max(0, this.getConnectionCheckIntervalMillis() - TimeUnit.NANOSECONDS.toMillis(status.elapsedTimeNano)));
+          //System.out.println("Delay " + delayMillis);
+          TimeUnit.MILLISECONDS.sleep(delayMillis);
```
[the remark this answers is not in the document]
sergiyvamz: Agree. I added a corresponding adjustment.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/NodeMonitoringConnectionPlugin.java
```diff
@@ -231,20 +304,7 @@ protected boolean doesNeedMonitoring(Class<?> methodInvokeOn, String methodName)
...
-
-    for (final String method : METHODS_EQUAL_TO) {
-      if (method.equals(methodName)) {
-        return false;
-      }
-    }
-
-    // Monitor all the other methods
-    return true;
+    return !METHODS_EQUAL_TO.contains(methodName);
```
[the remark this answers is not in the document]
sergiyvamz: I renamed variable to get it more clarity. As for method, it's a protected method and the idea was to have it protected and extendable in the future.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
```diff
@@ -139,42 +127,114 @@ public void stopMonitoring(MonitorConnectionContext context) {
...
+
+          if (delayMillis == -1) {
+            // No active contexts
+            delayMillis = THREAD_SLEEP_WHEN_INACTIVE_MILLIS;
+          } else {
+            delayMillis -= status.elapsedTimeNano;
+            // Check for min delay between node health check
+            if (delayMillis < MIN_CONNECTION_CHECK_TIMEOUT_MILLIS) {
+              delayMillis = MIN_CONNECTION_CHECK_TIMEOUT_MILLIS;
+            }
```
congoamz, line 229: should the elapsedTime subtraction come after this reset to the MIN_CONNECTION timeout? For example, if we spent 1s on the status check and the MIN_CONNECTION timeout is 3s, we only have to wait 2s

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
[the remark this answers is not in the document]
congoamz: Not sure if you misunderstood, I was wondering if `stopped` needs to be volatile since it is read by the main thread and written by the monitor thread. `contextLastUsedTimeStampNano` is as well, so it may need to be volatile too, I'm not sure though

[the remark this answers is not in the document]
sergiyvamz: Makes sense. Thank you.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/Monitor.java
sergiyvamz: I was thinking about this scenario but I guess it makes sense to avoid checking node status too often. So MIN_CONNECTION is applied to delayMillis.

congoamz (approved):

src/main/core-api/java/com/mysql/cj/util/CacheMap.java
```diff
@@ -0,0 +1,141 @@
...
+    return cacheItem == null ? null : cacheItem.item;
+  }
+
+  public V get(final K key, final V defaultItemValue, long itemExpirationNano) {
+    CacheItem<V> cacheItem = cache.compute(key,
+        (kk, vv) -> (vv == null || vv.isExpired())
+            ? new CacheItem<>(defaultItemValue, System.nanoTime() + itemExpirationNano)
+            : vv);
+    return cacheItem.item;
+  }
```
karenc-bq, line 59: I think these methods can be simplified if you add a getter for cacheItem.item:

```java
  public V getItem() {
      return isExpired() ? null : item;
    }
```

src/main/core-api/java/com/mysql/cj/util/CacheMap.java
```diff
@@ -0,0 +1,141 @@
...
+
+  private void cleanUp() {
+    if (this.cleanupTimeNanos.get() < System.nanoTime()) {
+      this.cleanupTimeNanos.set(System.nanoTime() + cleanupIntervalNanos);
+      cache.entrySet().forEach(entry -> {
+        if (entry.getValue() == null || entry.getValue().isExpired()) {
+          cache.remove(entry.getKey());
+        }
+      });
+    }
```
karenc-bq, line 90:
```suggestion
    if (this.cleanupTimeNanos.get() < System.nanoTime()) {
      this.cleanupTimeNanos.set(System.nanoTime() + cleanupIntervalNanos);
      cache.forEach((key, value) -> {
        if (value == null || value.isExpired()) {
          cache.remove(key);
        }
      });
    }
```

src/main/core-api/java/com/mysql/cj/util/CacheMap.java
```diff
@@ -0,0 +1,141 @@
...
+      }
+      CacheItem<?> other = (CacheItem<?>) obj;
+      if (item == null) {
+        if (other.item != null) {
+          return false;
+        }
+      } else if (!item.equals(other.item)) {
+        return false;
+      }
+      return true;
```
karenc-bq, line 133:
```suggestion
      if (item == null) {
        return other.item == null;
      } else {
        return item.equals(other.item);
      }
```

YoungHu:
this fix have bug will case #412

Issue #361, opened 2023-02-14
aws-jdbc-driver 1.1.4 + Hikari4.0.3 + AWSSecretsManagerPluginFactory - JDBC access denied


peter-jang:
### Describe the bug

Hi,
I'm trying to use AWS secret manager with HikariCP and aws-mysql-jdbc driver.
I followed document described here and it looks the initial JDBC connection working well.
However every 20minute or so, I get access denied error from my RDS Mysql error logs.
If I omit FailoverConnectionPluginFactory and NodeMonitoringConnectionPluginFactory from my application.property 
these error logs are not appear anymore.

I think that monitoring plugin could not get credential information which provided by AWS SecretManager.

When I start my application I see there's 10 connection made to RDS MySQL from my (ec2) server 
it is because HikariCP create 10 connection to manage connection pool.
While  1 connection out of 10 works well (I can query to RDS from my application),
I get 9 of "[Note] Access denied for user 'ec2-user'@'172.31.39.60' (using password: NO)" error 

### Expected Behavior

There should be no access denied error even though I used AWS secret manager with monitoring plugin.
If I turn off the enhanced monitoring feature (omit FailoverConnectionPluginFactory and NodeMonitoringConnectionPluginFactory from my application.property), I don't see access denied error anymore which is expected behavior.

### Current Behavior

every 20 min after starting application, I get access denied error from RDS error log.
Also, I was able to capture the access denied error via tcpdump as soon as the application was starting.

<img width="1156" alt="Screen Shot 2023-02-14 at 3 48 11 PM" src="https://user-images.githubusercontent.com/11946914/218660480-8e5653f5-4df0-4042-962d-c7da920b3b9e.png">

<img width="1508" alt="Screen Shot 2023-02-14 at 4 17 31 PM" src="https://user-images.githubusercontent.com/11946914/218666160-600ae7cb-b466-43bf-8c74-a3bc475361ea.png">

If I add below static credential information to application.property then error log disappears.
but I can't use this because I want to do credential rotation using AWS Secret Manager
```
spring.datasource.username=admin
spring.datasource.password=password-string
```

### Reproduction Steps

Create simple spring boot application with below configuration (I've used maven)
build project and run application, wait 20min to see RDS error logs.
or you can capture traffic dump using tcpdmp while application starting.
`tcpdump -nn -X port 3306`

**pom.xml**
```
                <dependency>
                         <groupId>software.aws.rds</groupId>
[...]

PR #369, opened 2023-03-08
fix: AWSSecretsManagerPlugin to update ConnectionProxy HostInfo on openInitialConnection (#361)

aaronchung-bitquill:
### Summary

fix: AWSSecretsManagerPlugin to update ConnectionProxy HostInfo on openInitialConnection (#361)

### Description

Plugins use the currentHostInfo from ConnectionProxy to create connections. Unfortunately, it doesnt get updated with credentials by AWSSecretsManagerPlugin, so connections made this way result in access denials. 

Changed AWSSecretsManagerPlugin to update the currentHostInfo of the CurrentConnectionProvider with user/password so that subsequent plugins that use for creating connections won't get "access denied". 

After talking with @sergiyv-improving , also disabled the code in FailoverConnectionPlugin in attemptConnectionUsingCachedTopology() method as it does not work with the AWSSecretsManagerPlugin, and has no proven improved performance benefits. 

### Additional Reviewers

@sergiyv-improving @karenc-bq @aaron-congo @joyc-bq @brunos-bq

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
@@ -169,7 +169,7 @@ public class ConnectionProxy implements ICurrentConnectionProvider, InvocationHa
   @Override
   public void setCurrentConnection(JdbcConnection connection, HostInfo info) {
     try {
-      if (this.currentConnection != null && !this.currentConnection.isClosed()) {
+      if (this.currentConnection != null && !this.currentConnection.equals(connection)  && !this.currentConnection.isClosed()) {
         this.currentConnection.close();
       }
     } catch (SQLException sqlEx) {
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
@@ -32,9 +32,10 @@
 package com.mysql.cj.jdbc.ha.plugins;
 
 import com.mysql.cj.conf.ConnectionUrl;
+import com.mysql.cj.conf.HostInfo;
 import com.mysql.cj.conf.PropertySet;
 import com.mysql.cj.exceptions.CJException;
-import com.mysql.cj.jdbc.ConnectionImpl;
+import com.mysql.cj.jdbc.ha.ConnectionUtils;
 import com.mysql.cj.log.Log;
 import com.mysql.cj.util.LRUCache;
 
@@ -67,7 +68,7 @@ public class AWSSecretsManagerPlugin implements IConnectionPlugin {
   static final String SQLSTATE_ACCESS_ERROR = "28000";
 
   static final LRUCache<Pair<String, Region>, Secret> SECRET_CACHE = new LRUCache<>(100);
-
+  private final ICurrentConnectionProvider currentConnectionProvider;
   private final IConnectionPlugin nextPlugin;
   private final Log logger;
   private final String secretId;
@@ -129,6 +130,7 @@ public class AWSSecretsManagerPlugin implements IConnectionPlugin {
           String.format("Configuration parameter '%s' is required.", REGION_PROPERTY));
     }
 
+    this.currentConnectionProvider = currentConnectionProvider;
     this.nextPlugin = nextPlugin;
     this.logger = logger;
     this.secretId = propertySet.getStringProperty(SECRET_ID_PROPERTY).getValue();
@@ -240,6 +242,13 @@ public class AWSSecretsManagerPlugin implements IConnectionPlugin {
     return fetched;
   }
 
+  private void updateCurrentHostInfoWithSecrets(HostInfo hostInfo) {
+    HostInfo currentHostInfo = this.currentConnectionProvider.getCurrentHostInfo();
+    this.currentConnectionProvider.setCurrentConnection(
+        this.currentConnectionProvider.getCurrentConnection(),
+        ConnectionUtils.copyWithAdditionalProps(currentHostInfo, hostInfo));
+  }
+
   /**
    * Called to open a new connection. This plugin is responsible to providing a recent credentials and delegate actual
    * opening a new connection to other plugins in the plugin chain. Eventually a new connection is handled either by
@@ -255,6 +264,7 @@ public class AWSSecretsManagerPlugin implements IConnectionPlugin {
         ConnectionUrl.getConnectionUrlInstance(connectionUrl.getDatabaseUrl(),
             props);
     this.nextPlugin.openInitialConnection(newConnectionUrl);
+    updateCurrentHostInfoWithSecrets(newConnectionUrl.getMainHost());
   }
 
   Secret fetchLatestCredentials() throws SecretsManagerException, JsonProcessingException {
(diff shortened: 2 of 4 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
```diff
@@ -169,7 +169,7 @@ public HostInfo getCurrentHostInfo() {
   @Override
   public void setCurrentConnection(JdbcConnection connection, HostInfo info) {
     try {
-      if (this.currentConnection != null && !this.currentConnection.isClosed()) {
+      if (this.currentConnection != null && !this.currentConnection.equals(connection)  && !this.currentConnection.isClosed()) {
```
brunos-bq, line 172: Not sure i understand. We close the current connection if we are setting a new connection as current?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -410,10 +410,11 @@ private void attemptConnectionUsingCachedTopology() throws SQLException {
 
     metricsContainer.registerUseCachedTopology(true);
 
-    int candidateIndex = getCandidateIndexForInitialConnection();
-    if (candidateIndex != NO_CONNECTION_INDEX) {
-      connectTo(candidateIndex);
-    }
+    // Disabling the following because it creates Access Denied when using AwsSecretsManagerPlugin and has no proven performance benefit
```
brunos-bq, line 413: Can you remove this code?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
aaronchung-bitquill: Yup, thats right. Its cleaning up the "old" current connection before we assign a new current connection. 
This fix here makes sure that if the new and old one the same, dont close it.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill: Will do. I'll update the comment  as well to say what the code was

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -410,10 +410,9 @@ private void attemptConnectionUsingCachedTopology() throws SQLException {
 
     metricsContainer.registerUseCachedTopology(true);
 
-    int candidateIndex = getCandidateIndexForInitialConnection();
-    if (candidateIndex != NO_CONNECTION_INDEX) {
-      connectTo(candidateIndex);
-    }
+    // Removed code to call getCandidateIndexForInitialConnection(), and connect to it. This is because on initial
```
karenc-bq, line 413: This method is called `attemptConnectionUsingCachedTopology`, and this method is called in `createConnection`. If we remove this piece of code it changes the intent of `attemptConnectionUsingCachedTopology`.
Do we still need `attemptConnectionusingCachedTopology`? And do we need to modify the code that is using this function?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
```diff
@@ -240,6 +242,12 @@ private boolean updateSecret(boolean forceReFetch) throws SQLException {
     return fetched;
   }
 
+  private void attemptToUpdateCurrentHostInfoWithSecrets(HostInfo hostInfo) {
+    HostInfo currentHostInfo = this.currentConnectionProvider.getCurrentHostInfo();
+    this.currentConnectionProvider.setCurrentConnection(this.currentConnectionProvider.getCurrentConnection(),
+        ConnectionUtils.copyWithAdditionalProps(currentHostInfo, hostInfo));
```
karenc-bq, line 248: super nit:
If we are breaking the parameters due to line limit, I prefer we put each parameter on a new line, i.e.
```java
    this.currentConnectionProvider.setCurrentConnection(
        this.currentConnectionProvider.getCurrentConnection(),
        ConnectionUtils.copyWithAdditionalProps(currentHostInfo, hostInfo));
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill: Good catch! I think we should remove the method as it also has the effect of updating the FailoverConnectionPlugin hosts with the cached hosts. But rename it seems like a good idea.

aaronchung-bitquill: I just noticed that this also leaves some unused private methods. I'll remove those as well.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill: I just noticed that this also leaves some unused private methods. I'll remove those as well.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
```diff
@@ -240,6 +242,12 @@ private boolean updateSecret(boolean forceReFetch) throws SQLException {
     return fetched;
   }
 
+  private void attemptToUpdateCurrentHostInfoWithSecrets(HostInfo hostInfo) {
```
congoamz, line 245: super nit: I think just `updateCurrentHostInfoWithSecrets` will work for this method name

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
aaronchung-bitquill: Done!

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/AWSSecretsManagerPlugin.java
aaronchung-bitquill: Done!

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill: Removed commented out code and comments

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill: Update: Leaving in the method, but with the name changed to reflect what the method actually does now. Also removed SqlException from the signature, and try catch from the user of this method. .

karenc-bq (approved):

PR #373, opened 2023-03-15
refactor: reduce map lookups to reduce number of locks required

karenc-bq:
### Summary

Instead of calling `containsKey` then calling `get`, which requires looking up the map twice, simply call get and check the value. This works because the map doesn't store null values.

### Additional Reviewers

@brunos-bq
@aaronchung-bitquill

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
@@ -61,6 +61,7 @@ import javax.net.ssl.SSLException;
 import java.io.EOFException;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -92,22 +95,22 @@ public class FailoverConnectionPlugin implements IConnectionPlugin {
...
-      "execute",
-      "executeBatch",
-      "executeLargeBatch",
-      "executeLargeUpdate",
-      "executeQuery",
-      "executeUpdate",
-      "executeWithFlags",
-      "getParameterMetaData"
-  ));
+  private final static Set<String> METHODS_REQUIRE_UPDATED_TOPOLOGY = new ConcurrentHashMap<String, String>() {{
```
brunos-bq, line 98: will you use any Map functions on this Map? Or just searches on the keySet?

sergiyvamz: ConcurrentHashMap.newKeySet()

sergiyvamz (approved):

Issue #407, opened 2023-05-12
all connections reference to different databases of the same rds cluster connected to the same database after failover


little-apppple:
### Describe the bug

there are some problems while the we are using multi databases in the same db cluster in our application. All of the database will change to the the same one after rds failover, for expamle  when  my application configed 
`jdbc:mysql:aws://example.rds.amazonaws.com:3306/database_00?useUnicode=true&character_set_server=utf8mb4&characterEncoding=utf-8&useSSL=false` 
 `jdbc:mysql:aws://example.rds.amazonaws.com:3306/database_01?useUnicode=true&character_set_server=utf8mb4&characterEncoding=utf-8&useSSL=false`
 `jdbc:mysql:aws://example.rds.amazonaws.com:3306/database_02?useUnicode=true&character_set_server=utf8mb4&characterEncoding=utf-8&useSSL=false`
  `jdbc:mysql:aws://example.rds.amazonaws.com:3306/database_03?useUnicode=true&character_set_server=utf8mb4&characterEncoding=utf-8&useSSL=false`
  and then 
  `AuroraTopologyService.topologyCache` cached 'example.rds.amazonaws.com:3306'   originalUrl only contains jdbc:mysql:aws://example.rds.amazonaws.com:3306/database_03?useUnicode=true&character_set_server=utf8mb4&characterEncoding=utf-8&useSSL=false .

All of those connections will reference to the database of 'database_03' after failover

### Expected Behavior

all connections of this rds cluster connected to the origin database of each after failover

### Current Behavior

all connections of this rds cluster connected  to the same database after failover

### Reproduction Steps

1.config multi databases of one rds cluster in your application
2.rds cluster failover
3.write data to different databases in your application

### The AWS JDBC Driver for MySQL version used

1.1.6

### JDK version used

1.8.0_181

### Operating System and version

container: alpine-release:3.16.0

YoungHu:
we has same issue. root cause is cluster topology is shared with clusterId and topology contains database information such as endpoint_url、db_name、username、password. in multi datasource use same cluster case, when failover recreate connection will use the topology HostInfo direct, that will cause re-conn to other database. the solution is when recreate connection, need use current connection info(db_name, username, password, db_properties) to replace topology info.
I just verify and fix `ClusterAwareWriterFailoverHandler`, hope official bug fix.

[...]

PR #408, opened 2023-05-15
fix issue#407

YoungHu:
### Summary

fix issue #407 

### Description

when failover re-create connection, replace dbname, username, password with current connection instand of  topology hostInfo

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionUtils.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionUtils.java
@@ -34,7 +34,7 @@ package com.mysql.cj.jdbc.ha;
 import com.mysql.cj.conf.ConnectionUrl;
 import com.mysql.cj.conf.DatabaseUrlContainer;
 import com.mysql.cj.conf.HostInfo;
-import com.mysql.cj.jdbc.JdbcConnection;
+import com.mysql.cj.conf.PropertyKey;
 
 import java.sql.SQLException;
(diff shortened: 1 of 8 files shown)


crystall-bitquill:
Hi @YoungHu,

Thank you for your PR and the effort you've spent making improvements for the driver! I recently made a [PR to address the same issue](https://github.com/awslabs/aws-mysql-jdbc/pull/410), and it takes a different approach. You'll been tagged in the PR for visibility, but the reasoning we took a different approach was:
- To avoid passing any connection objects to the failover handlers and therefore through additional methods
- To address any connection properties other than the database name that may have been retrieved incorrectly from the topology

Thank you again for your feedback in identifying the improvement, and please feel free to make contributions again in the future!

YoungHu:
> Hi @YoungHu,
> 
> Thank you for your PR and the effort you've spent making improvements for the driver! I recently made a [PR to address the same issue](https://github.com/awslabs/aws-mysql-jdbc/pull/410), and it takes a different approach. You'll been tagged in the PR for visibility, but the reasoning we took a different approach was:
> 
> * To avoid passing any connection objects to the failover handlers and therefore through additional methods
> * To address any connection properties other than the database name that may have been retrieved incorrectly from the topology
> 
> Thank you again for your feedback in identifying the improvement, and please feel free to make contributions again in the future!

@crystall-bitquill wow, it's seems a better solution to resolve this issue with easy way. we will use ourself patch to fix affected application first, hope official snapshot/release ASAP.

PR #410, opened 2023-05-18
refactor: remove host properties from topology

crystall-bitquill:
### Summary

Remove additional host properties from topology

### Description

When the failover plugin is initialized, it passes properties to the AuroraTopologyService. This means that when the topology service gets a topology, every HostInfo object in the topology will contain the same properties. The topology is stored in a cache, which is identified by the cluster ID. This means different connections to the same cluster may retrieve topology information containing properties of each other.

This PR removes the host properties other than the host and port number from the topology service and instead sets the initialConnectionProps to contain them instead.

Addresses issue #407 

### Additional Reviewers

@YoungHu

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionUtils.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionUtils.java
@@ -109,11 +109,11 @@ public class ConnectionUtils {
   public static HostInfo copyWithAdditionalProps(
       HostInfo baseHostInfo,
       HostInfo newHostInfo) {
-    DatabaseUrlContainer urlContainer = ConnectionUrl.getConnectionUrlInstance(
+    final DatabaseUrlContainer urlContainer = ConnectionUrl.getConnectionUrlInstance(
         baseHostInfo.getDatabaseUrl(),
         new Properties());
-    Map<String, String> originalProps = baseHostInfo.getHostProperties();
-    Map<String, String> mergedProps = new HashMap<>();
+    final Map<String, String> originalProps = baseHostInfo.getHostProperties();
+    final Map<String, String> mergedProps = new HashMap<>();
     mergedProps.putAll(originalProps);
     mergedProps.putAll(newHostInfo.getHostProperties());
 
@@ -123,16 +123,16 @@ public class ConnectionUtils {
   }
 
   public static HostInfo createHostWithProperties(HostInfo baseHost, Map<String, String> properties) {
-    Map<String, String> propertiesCopy = new HashMap<>(properties);
+    final Map<String, String> propertiesCopy = new HashMap<>(properties);
     propertiesCopy.putAll(baseHost.getHostProperties());
-    String hostEndpoint = baseHost.getHost();
-    int port = baseHost.getPort();
-    String user = propertiesCopy.get(PropertyKey.USER.getKeyName());
-    String password = propertiesCopy.get(PropertyKey.PASSWORD.getKeyName());
+    final String hostEndpoint = baseHost.getHost();
+    final int port = baseHost.getPort();
+    final String user = propertiesCopy.get(PropertyKey.USER.getKeyName());
+    final String password = propertiesCopy.get(PropertyKey.PASSWORD.getKeyName());
     propertiesCopy.remove(PropertyKey.USER.getKeyName());
     propertiesCopy.remove(PropertyKey.PASSWORD.getKeyName());
 
-    ConnectionUrl hostUrl = ConnectionUrl.getConnectionUrlInstance(
+    final ConnectionUrl hostUrl = ConnectionUrl.getConnectionUrlInstance(
         getUrlFromEndpoint(
             hostEndpoint,
             port),
@@ -145,16 +145,6 @@ public class ConnectionUtils {
         user,
         password,
         propertiesCopy);
-  }
-
-  public static List<HostInfo> createTopologyFromSimpleHosts(List<HostInfo> basicTopology, Map<String, String> properties) {
-    List<HostInfo> topology = new ArrayList<>();
-    if (basicTopology != null) {
-      for (HostInfo host : basicTopology) {
-        topology.add(createHostWithProperties(host, properties));
-      }
-    }
-    return topology;
   }
 
   private static String getUrlFromEndpoint(String endpoint, int port) {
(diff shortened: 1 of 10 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionUtils.java
```diff
@@ -119,6 +122,49 @@ public static HostInfo copyWithAdditionalProps(
...
+    String user = propertiesCopy.get(PropertyKey.USER.getKeyName());
+    String password = propertiesCopy.get(PropertyKey.PASSWORD.getKeyName());
+    propertiesCopy.remove(PropertyKey.USER.getKeyName());
+    propertiesCopy.remove(PropertyKey.PASSWORD.getKeyName());
+
+    ConnectionUrl hostUrl = ConnectionUrl.getConnectionUrlInstance(
+        getUrlFromEndpoint(
+            hostEndpoint,
+            port),
+        new Properties());
```
karenc-bq, line 139: Why are you using an empty properties here

crystall-bitquill: Before in the AuroraTopologyService, host URLs were created with empty properties, so I kept that behaviour here.

src/test/java/com/mysql/cj/jdbc/ha/plugins/failover/ClusterAwareTestUtils.java
```diff
@@ -56,12 +62,58 @@ protected static HostInfo createBasicHostInfo(
...
+  protected static class HostInfoMatcher implements ArgumentMatcher<HostInfo> {
+
+    private final HostInfo expectedHost;
+
+    HostInfoMatcher(HostInfo hostInfo) {
+      this.expectedHost = hostInfo;
+    }
+
+    @Override
+    public boolean matches(HostInfo host) {
```
karenc-bq, line 87: Can you override equals here instead

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -1334,14 +1314,20 @@ private void validateHostPatternSetting(String hostPattern) throws SQLException
...
     final Map<String, String> initialConnectionProperties = new HashMap<>();
     final Properties originalProperties = propertySet.exposeAsProperties();
     originalProperties.stringPropertyNames()
         .stream()
         .filter(x -> this.propertySet.getProperty(x).isExplicitlySet())
         .forEach(x -> initialConnectionProperties.put(x, originalProperties.getProperty(x)));
 
+    initialConnectionProperties.putAll(currentHost.getHostProperties());
+    initialConnectionProperties.put(PropertyKey.USER.getKeyName(), currentHost.getUser());
+    initialConnectionProperties.put(PropertyKey.PASSWORD.getKeyName(), currentHost.getPassword());
```
congoamz, line 1327: Would we ever run into the situation where the user/password are set in the current host properties but getUser/getPassword return null?

src/test/java/com/mysql/cj/jdbc/ha/plugins/failover/ClusterAwareTestUtils.java
crystall-bitquill: No, this is an ArgumentMatcher method.

src/test/java/com/mysql/cj/jdbc/ha/plugins/failover/ClusterAwareTestUtils.java
```diff
@@ -56,12 +62,52 @@ protected static HostInfo createBasicHostInfo(
...
+    boolean sameProperties = hostInfo1.getHostProperties().size() == hostInfo2.getHostProperties().size();
+    for (Map.Entry<String,String> property : hostInfo1.getHostProperties().entrySet()) {
+      if (hostInfo2.getHostProperties().get(property.getKey()) == null ||
+        !Objects.equals(
+            hostInfo2.getHostProperties().get(property.getKey()),
+            hostInfo1.getHostProperties().get(property.getKey()))) {
+        sameProperties = false;
+        break;
+      }
+    }
```
karenc-bq, line 105:
```suggestion
    final boolean sameProperties = Objects.equals(hostInfo1.exposeAsProperties(), hostInfo2.exposeAsProperties());
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/AuroraTopologyService.java
```diff
@@ -165,8 +163,7 @@ public void setClusterInstanceTemplate(HostInfo clusterInstanceTemplate) {
     this.log.logTrace(Messages.getString(
         "AuroraTopologyService.2",
         new Object[] {clusterInstanceTemplate.getHost(),
-            clusterInstanceTemplate.getPort(),
```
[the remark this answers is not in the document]
crystall-bitquill: The database was removed from the instance template because it's one of the properties that shouldn't be shared between connections to the same cluster. Good catch with the error message though

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/ClusterAwareReaderFailoverHandler.java
```diff
@@ -193,7 +193,9 @@ private Future<ReaderFailoverResult> submitInternalFailoverTask(
 
             // need to ensure that the new connection is a connection to a reader node
             final HostInfo newHost = hosts.get(result.getConnectionIndex());
-            final List<HostInfo> topology = this.topologyService.getTopology(result.getConnection(), true);
+            final List<HostInfo> topology = ConnectionUtils.createTopologyFromHosts(
```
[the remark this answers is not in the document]
crystall-bitquill: I've updated the PR so this method isn't needed. Instead, any host properties are added when connections are made. Does that work better?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
crystall-bitquill: I don't think so, the driver checks host properties for the user/password before this point and will use the default user/password if nothing is there.

karenc-bq (approved):

sergiyvamz (approved):

PR #413, opened 2023-06-06
fix: Accept "convertToNull" value for property "zeroDateTimeBehaviour"

crystall-bitquill:
### Summary

Accept "convertToNull" value for property "zeroDateTimeBehaviour"

### Description

Some connection properties such as `zeroDateTimeBehaviour` are altered in the `fixHostInfo` method. The `zeroDateTimeBehaviour` value in particular is replaced in the `replaceLegacyPropertyValues` method. This allows the driver to accept both upper and lowercase values. When the driver detects an AWS JDBC Driver protocol, it will create a proxy using the properties the user specified and not the corrected set. This PR will use the "fixed" properties instead.

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
@@ -54,6 +54,7 @@ import java.lang.reflect.Proxy;
 import java.sql.SQLException;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Properties;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/ConnectionProxy.java
```diff
@@ -225,7 +226,13 @@ protected void initLogger(ConnectionUrl connUrl) {
...
   protected void initSettings(ConnectionUrl connectionUrl) throws SQLException {
     try {
-      this.connProps.initializeProperties(connectionUrl.getConnectionArgumentsAsProperties());
+      Properties props = new Properties();
+      for (Map.Entry<String, String> entry : connectionUrl.getMainHost().getHostProperties().entrySet()) {
+        if (entry.getValue() != null) {
+          props.put(entry.getKey(), entry.getValue());
+        }
+      }
+      this.connProps.initializeProperties(props);
```
karenc-bq, line 235:
```suggestion
      final Properties props = new Properties();
      props.putAll(connectionUrl.getMainHost().getHostProperties());
      this.connProps.initializeProperties(props);
```

crystall-bitquill: This will cause a null pointer exception if one of the properties is null.

karenc-bq (approved):

Issue #303, opened 2022-10-27
Several re-connection issues with Aurora v2


albanf:
### Describe the bug

After a failover, some connections are invalid:
* they are connected to the replica instead of the writer
* the failover is disabled

### Expected Behavior

Connections to writer should reconnect to the new writer, and the failover should still be enabled

### Current Behavior

After a failover, some connections are invalid:
* they are connected to the replica instead of the writer
* the failover is disabled

### Reproduction Steps

It happens with multiple connections only, because some go through the failover process, which updates the topology in the shared cache.
Then other connections are using incorrectly the new topology to update actual connection.

### Possible Solution

PR: https://github.com/awslabs/aws-mysql-jdbc/pull/302

### The AWS JDBC Driver for MySQL version used

1.1.1

### JDK version used

11

### Operating System and version

Alpine 3.16

karenc-bq:
Ongoing discussion in PR #302

amka-ar:
Hello. I want to raise this issue.

First, by explicitly invoking `setReadOnly` each time we get a connection from the pool, we were able to [auto-recover](https://github.com/awslabs/aws-mysql-jdbc/blob/main/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java#L1205) from most `The MySQL server is running with the --read-only option so it cannot execute this statement` cases we have had. Looks like this behavior is not documented.

But even with this feature we got one unrecoverable case and we believe the cause was the same as [here](https://github.com/awslabs/aws-mysql-jdbc/pull/400#issuecomment-1530850280) (two records with `MASTER_SESSION_ID` during the failover and `isMultiWriterCluster`). In 1.1.7 `isMultiWriterCluster` was removed, but at the same time, `AuroraTopologyService` started removing the oldest `MASTER_SESSION_ID` record (previous writer) from the list of hosts. And that brings us to this issue.

[...]

PR #426, opened 2023-06-27
fix: keep currentHostIndex and hosts list in sync and pick new connection when host role changes

crystall-bitquill:
### Summary

Keep currentHostIndex and hosts list in sync and pick new connection when host role changes

### Description

When picking a new connection, the failover plugin should use the latest topology if it is available. A check was also added to ensure that when the topology is updated, if the current host changes to a reader, a new connection is picked.

Addresses #303.

### Additional Reviewers

@albanf

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
@@ -174,6 +174,7 @@ public class FailoverConnectionPlugin implements IConnectionPlugin {
 
   private long invokeStartTimeMs;
   private long failoverStartTimeMs;
+  boolean isInitialConnectionToReader;
 
   public FailoverConnectionPlugin(
       ICurrentConnectionProvider currentConnectionProvider,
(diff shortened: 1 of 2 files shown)


src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -174,6 +174,7 @@ public class FailoverConnectionPlugin implements IConnectionPlugin {
 
   private long invokeStartTimeMs;
   private long failoverStartTimeMs;
+  boolean initialConnectionToReader;
```
aaronchung-bitquill, line 177: Nit - The naming of this boolean variable could be more concise and named in a form of a statement. For example `isInitialConnectionToReader`. Otherwise, without looking at the type, it could be confused as a connection object.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
```diff
@@ -618,6 +619,15 @@ protected void invalidateCurrentConnection() {
   }
 
   protected synchronized void pickNewConnection() throws SQLException {
+    pickNewConnection(null);
+  }
+
+  synchronized void pickNewConnection(List<HostInfo> latestTopology) throws SQLException {
+    List<HostInfo> currentTopology = this.hosts;
```
karenc-bq, line 626:
```suggestion
    final List<HostInfo> currentTopology = this.hosts;
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/failover/FailoverConnectionPlugin.java
aaronchung-bitquill, line 177: Also just noticed, should this field be `private`?

crystall-bitquill: It's used in a few tests, so I made it package private.

karenc-bq (approved):

Issue #464, opened 2023-09-25
During batchUpdate() Single line comments are causing error "Statement.executeUpdate() or Statement.executeLargeUpdate() cannot issue statements that produce result sets."


benkanlu:
### Describe the bug

Driver does not properly deal with single-line quotes when using batchUpdate with multiple statements that have various combinations of single-line quotes using '-- ' or '\n-- '. This bug appears to have been introduced in version 1.0.0

Usage of comments in SQL is very common and batchUpdate should be able to handle this correctly in various combinations.

### Expected Behavior

Similar to MariaDB/MySQL JDBC drivers, all comments (single-line, multi-line) should be properly ignored by the driver and not result in an confusing '... cannot issue statements that produce results sets.' error. 

### Current Behavior

2023-09-25T13:45:29.423-06:00  INFO 5952 --- [           main] c.e.B.BatchUpdateIssueApplication        : Starting BatchUpdateIssueApplication using Java 17.0.6 with PID 5952 (C:\xxxxxx\BatchUpdateIssue\BatchUpdateIssue\target\classes started by xxxxxxin C:\xxxxxx\BatchUpdateIssue\BatchUpdateIssue)
2023-09-25T13:45:29.425-06:00  INFO 5952 --- [           main] c.e.B.BatchUpdateIssueApplication        : No active profile set, falling back to 1 default profile: "default"
2023-09-25T13:45:29.772-06:00  INFO 5952 --- [           main] c.e.B.BatchUpdateIssueApplication        : Started BatchUpdateIssueApplication in 0.525 seconds (process running for 0.882)
2023-09-25T13:45:29.775-06:00  INFO 5952 --- [           main] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Starting...
2023-09-25T13:45:30.731-06:00  INFO 5952 --- [           main] com.zaxxer.hikari.pool.HikariPool        : HikariPool-1 - Added connection software.aws.rds.jdbc.mysql.shading.com.mysql.cj.jdbc.ConnectionImpl@508a65bf
2023-09-25T13:45:30.732-06:00  INFO 5952 --- [           main] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Start completed.
2023-09-25T13:45:31.381-06:00  INFO 5952 --- [           main] .s.b.a.l.ConditionEvaluationReportLogger : 

Error starting ApplicationContext. To display the condition evaluation report re-run your application with 'debug' enabled.
[...]

PR #470, opened 2023-10-05
fix: batch execute with comments results in error

crystall-bitquill:
### Summary

Batch execute with comments results in error

### Description

Batch execute with comments resulted in errors about being unable to issue statements that produce result sets. The `isNonResultSetProducingQuery` method was missing the `None` query return type.

Addresses #464

### DIFF
--- a/src/test/java/testsuite/regression/StatementRegressionTest.java
+++ b/src/test/java/testsuite/regression/StatementRegressionTest.java
@@ -12707,4 +12707,46 @@ public class StatementRegressionTest extends BaseTestCase {
             }
         }
     }
+
+    /**
+     * Test fix for Issue #464. Executing a batch statement should ignore any comments.
+     * Verifies that isNonResultSetProducingQuery() properly handles queries that return
+     * queryReturnType == QueryReturnType.NONE
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testBugIssue464() throws Exception {
+        Properties props = new Properties();
+        props.setProperty(PropertyKey.sslMode.getKeyName(), SslMode.DISABLED.name());
+        props.setProperty(PropertyKey.allowPublicKeyRetrieval.getKeyName(), "true");
+        this.conn = getConnectionWithProps(props);
+
+        String tableName = "testBugIssue464";
+
+        final Statement statement = this.conn.createStatement();
+        createTable(tableName, "(c0 INT)");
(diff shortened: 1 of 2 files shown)


src/test/java/testsuite/regression/StatementRegressionTest.java
```diff
@@ -12707,4 +12707,44 @@ public void testBugIssue450() throws Exception {
             }
         }
     }
+
+    /**
+     * Test fix for Issue #464. Executing a batch statement should ignore any comments.
+     *
```
brunos-bq, line 12713: NIT: I'd suggest adding an extra line like
"Verifying that isNonResultSetProducingQuery() properly handles queries that return queryReturnType == QueryReturnType.NONE

src/test/java/testsuite/regression/StatementRegressionTest.java
```diff
@@ -12707,4 +12707,44 @@ public void testBugIssue450() throws Exception {
...
+        Statement bstmt = conn.createStatement();
+        bstmt.addBatch("INSERT INTO " + tableName + " /* Comment 1 */ VALUES (1)");
+        bstmt.addBatch("INSERT INTO " + tableName + " VALUES (2)" + " # Comment 2");
+        bstmt.addBatch("INSERT INTO " + tableName + " VALUES (3)" + " -- Comment 3");
+        bstmt.addBatch("/* Comment 4 */");
+        bstmt.addBatch("# Comment 5");
+        bstmt.addBatch("-- Comment 6");
+
+        try {
+            bstmt.executeBatch();
```
brunos-bq, line 12737: is addBatch() that returns NONE or executeBatch() or both?

crystall-bitquill: The executeBatch() method triggers the query type check, addBatch() will not.

brunos-bq (approved):

karenc-bq (approved):

benkanlu:
Yes, that did seem to fix the problems with .batchUpdate() not handling comments correctly. Thank you. I would suggest  also opening a PR to the Oracle MySQL repo (https://github.com/mysql/mysql-connector-j) so that your fix will eventually get applied there. But, I guess if no additional work is going to be done on this driver for AWS, no point.  Thank you, regardless!

PR #474, opened 2023-10-13
fix: query timeout kill query directed to incorrect instance

crystall-bitquill:
### Summary

fix: query timeout kill query directed to incorrect instance

### Description

When a query timeout is set and the timeout value has passed, the driver sends a kill query to stop the process through a new connection that was created using the same host as the original connection. When using the reader cluster url, connections are randomly sent to one of the available reader instances. This will occasionally cause errors when the kill query is sent to a different instance where the statement was not executed, resulting in an "Unknown process id" message.

This PR adds a check in the default connection plugin for connections using the reader cluster url. Once the connection is made using the reader cluster url, a new connection is made to a reader instance that is obtained by querying for the current instance name. The original connection using the reader cluster is replaced with the connection to the instance.

Addresses #461

### DIFF
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/DefaultConnectionPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/DefaultConnectionPlugin.java
@@ -100,8 +100,9 @@ public class DefaultConnectionPlugin implements IConnectionPlugin {
       return;
     }
 
-    HostInfo mainHostInfo = connectionUrl.getMainHost();
+    final HostInfo mainHostInfo = connectionUrl.getMainHost();
     JdbcConnection connection = this.connectionProvider.connect(mainHostInfo);
+
     this.currentConnectionProvider.setCurrentConnection(connection, mainHostInfo);
   }
 
--- a/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
+++ b/src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License, version 2.0
+ * (GPLv2), as published by the Free Software Foundation, with the
+ * following additional permissions:
+ *
+ * This program is distributed with certain software that is licensed
+ * under separate terms, as designated in a particular file or component
+ * or in the license documentation. Without limiting your rights under
+ * the GPLv2, the authors of this program hereby grant you an additional
+ * permission to link the program and your derivative works with the
+ * separately licensed software that they have included with the program.
+ *
+ * Without limiting the foregoing grant of rights under the GPLv2 and
+ * additional permission as to separately licensed software, this
+ * program is also subject to the Universal FOSS Exception, version 1.0,
+ * a copy of which can be found along with its FAQ at
+ * http://oss.oracle.com/licenses/universal-foss-exception.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License, version 2.0, for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see
+ * http://www.gnu.org/licenses/gpl-2.0.html.
+ */
+
+package com.mysql.cj.jdbc.ha.plugins;
+
+import com.mysql.cj.conf.ConnectionUrl;
+import com.mysql.cj.conf.HostInfo;
+import com.mysql.cj.conf.PropertyKey;
+import com.mysql.cj.jdbc.JdbcConnection;
+import com.mysql.cj.jdbc.ha.util.ConnectionUtils;
+import com.mysql.cj.jdbc.ha.util.RdsUtils;
+import com.mysql.cj.util.StringUtils;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.concurrent.Callable;
+
+/**
+ * This connection plugin is used when connecting through reader cluster endpoints, where all new
+ * connections should be directed to the same reader instance rather than a random reader instance
+ * for each new connection.
+ */
+public class ReaderClusterConnectionPlugin implements IConnectionPlugin {
+  private static final String GET_INSTANCE_QUERY = "SELECT @@aurora_server_id";
+  protected IConnectionProvider connectionProvider;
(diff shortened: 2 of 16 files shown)


src/test/java/testsuite/integration/container/AuroraMysqlIntegrationTest.java
```diff
@@ -607,4 +618,16 @@ public void test_PreparedStatementHashCodes() throws SQLException, IOException {
     conn.close();
   }
 
+  @RepeatedTest(100)
```
sergiyvamz, line 621: Why do we need to run it so many times?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/DefaultConnectionPlugin.java
```diff
@@ -100,8 +107,24 @@ public void openInitialConnection(ConnectionUrl connectionUrl) throws SQLExcepti
...
+          rdsUtils.getRdsInstanceHostPattern(mainHostInfo.getHost()).replace("?", connectedHostName);
+      if (!StringUtils.isNullOrEmpty(mainHostInfo.getHostProperties().get(PropertyKey.clusterInstanceHostPattern.getKeyName()))) {
+        instanceEndpoint = mainHostInfo.getHostProperties().get(PropertyKey.clusterInstanceHostPattern.getKeyName())
+                .replace("?", connectedHostName);
+      }
+      final HostInfo instanceHostInfo = ConnectionUtils.createInstanceHostWithProperties(instanceEndpoint, mainHostInfo);
+      final JdbcConnection hostConnection = this.connectionProvider.connect(instanceHostInfo);
+      connection.close();
+      connection = hostConnection;
+    }
```
sergiyvamz, line 126: I'd suggest to relocate this logic into a new plugin. In this way, it'd be easier for users to use it when it's needed, or skip it. That gives more flexibility for users who don't experience any issue with query timeout.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/RdsUtils.java
```diff
@@ -0,0 +1,142 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License, version 2.0, for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see
+ * http://www.gnu.org/licenses/gpl-2.0.html.
+ */
+
+package com.mysql.cj.jdbc.ha;
```
sergiyvamz, line 32: com.mysql.cj.jdbc.ha.util ?

sergiyvamz: And similar suggestion's for the ConnectionUtils class above.

src/test/java/testsuite/integration/container/AuroraMysqlIntegrationTest.java
crystall-bitquill: When the new connection for the query timeout is made, the reader cluster endpoint doesn't always choose a reader different from the original connection so it's hard to reproduce this issue consistently.

sergiyvamz (approved):

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+ * the GPLv2, the authors of this program hereby grant you an additional
+ * permission to link the program and your derivative works with the
+ * separately licensed software that they have included with the program.
+ *
+ * Without limiting the foregoing grant of rights under the GPLv2 and
+ * additional permission as to separately licensed software, this
+ * program is also subject to the Universal FOSS Exception, version 1.0,
+ * a copy of which can be found along with its FAQ at
+ * http://oss.oracle.com/licenses/universal-foss-exception.
+ *
```
karenc-bq, line 21: Part of the license is duplicated.

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+         logger,
+         new BasicConnectionProvider());
+  }
+
+  public ReaderClusterConnectionPlugin(
+      ICurrentConnectionProvider currentConnectionProvider,
+      IConnectionPlugin nextPlugin,
+      Log logger,
+      IConnectionProvider connectionProvider) {
+    if (logger == null) {
```
karenc-bq, line 75: I don't think logger is used in this file. Do we need to store it?

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+        final String connectedHostName = getCurrentlyConnectedInstance(currentConnection);
+        String instanceEndpoint =
+            rdsUtils.getRdsInstanceHostPattern(mainHostInfo.getHost()).replace("?", connectedHostName);
+        if (!StringUtils.isNullOrEmpty(
+              mainHostInfo.getHostProperties().get(PropertyKey.clusterInstanceHostPattern.getKeyName()))) {
+          instanceEndpoint = mainHostInfo.getHostProperties().get(PropertyKey.clusterInstanceHostPattern.getKeyName())
+              .replace("?", connectedHostName);
+        }
+        final HostInfo instanceHostInfo =
+            ConnectionUtils.createInstanceHostWithProperties(instanceEndpoint, mainHostInfo);
```
karenc-bq, line 120:
```suggestion
        final String pattern = mainHostInfo.getHostProperties().get(PropertyKey.clusterInstanceHostPattern.getKeyName());
        final String instanceEndpoint = !StringUtils.isNullOrEmpty(pattern) ? pattern : rdsUtils.getRdsInstanceHostPattern(mainHostInfo.getHost());
        final HostInfo instanceHostInfo =
            ConnectionUtils.createInstanceHostWithProperties(
                instanceEndpoint.replace("?", connectedHostName), 
                mainHostInfo);
```

src/main/user-impl/java/com/mysql/cj/jdbc/ha/plugins/ReaderClusterConnectionPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+  @Override
+  public void transactionCompleted() {
+    this.nextPlugin.transactionCompleted();
+  }
+
+  private String getCurrentlyConnectedInstance(JdbcConnection connection) throws SQLException {
+    try (final Statement statement = connection.createStatement()) {
+      final ResultSet rs = statement.executeQuery(GET_INSTANCE_QUERY);
+      rs.next();
+      return rs.getString(1);
```
karenc-bq, line 146:
```suggestion
      if (rs.next()) {
        return rs.getString(1);
      }

      return null;
    }
```

karenc-bq (approved):

Issue #484, opened 2023-10-30
Statement.getGeneratedKeys() returns unexpected value


dwenking:
### Describe the bug

In my test case, when I use Statement to execute and then attempt to retrieve the latest GeneratedKeys, I expect to get a value of 3. 
However, the returned values are 1 and 2, which are the values I inserted previously. I think this behavior is incorrect. I also have this test case on MariaDB Connector J, and it returns 3. I think 3 may be more reasonable.

### Expected Behavior

See bug description.

### Current Behavior

See bug description.

### Reproduction Steps

```
@Test
public void test() throws SQLException {
    Connection con = DriverManager.getConnection("jdbc:mysql://localhost:3306/test?user=user&password=password");
    Statement stmt = con.createStatement();
    stmt.addBatch("DROP TABLE IF EXISTS table0_0;");
    stmt.addBatch("CREATE TABLE table0_0(id INT AUTO_INCREMENT PRIMARY KEY,value INT);");
    stmt.addBatch("INSERT INTO table0_0 VALUES(1, -179653912)");
    stmt.addBatch("INSERT INTO table0_0 VALUES(2, 1207965915)");
    stmt.executeBatch();

    stmt.executeUpdate("INSERT INTO table0_0 (value) VALUES(667711856)", Statement.RETURN_GENERATED_KEYS);
    ResultSet rs = stmt.getGeneratedKeys();
    while (rs.next()) {
        System.out.println(rs.getInt(1));
    }
}
```

### The AWS JDBC Driver for MySQL version used

1.1.10

### JDK version used

20

### Operating System and version

Windows

brunos-bq:
Hi @dwenking,

Thanks for reaching out and raising this issue.

We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

crystall-bitquill:
Hi @dwenking,

A fix for this issue (PR #490) was merged recently and was included in the latest release of the driver, version [1.1.11](https://search.maven.org/artifact/software.aws.rds/aws-mysql-jdbc/1.1.11/jar?eh=). Could you kindly check it out and let us know if the issue persists?

Thank you!

crystall-bitquill:
Hi @dwenking,

I wanted to check in and see if you were able to verify that the fix resolved the issue. If there are no further updates on this ticket in the next few days, it will be closed. However, if you have other questions or concerns, please feel free to reach out again.

Thank you!

PR #490, opened 2023-11-01
fix: incorrect generated keys after batch and individual statement executions

crystall-bitquill:
### Description

When batch statements are executed, followed by individual statements, the generated keys will be for the batch statements.  This PR allows the generated keys for the statement to be for the latest statement execution.

Addresses #484

### DIFF
--- a/src/test/java/testsuite/regression/StatementRegressionTest.java
+++ b/src/test/java/testsuite/regression/StatementRegressionTest.java
@@ -12724,7 +12724,6 @@ public class StatementRegressionTest extends BaseTestCase {
 
         String tableName = "testBugIssue464";
 
-        final Statement statement = this.conn.createStatement();
         createTable(tableName, "(c0 INT)");
 
         Statement bstmt = conn.createStatement();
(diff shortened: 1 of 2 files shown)


src/test/java/testsuite/regression/StatementRegressionTest.java
```diff
@@ -12724,7 +12724,6 @@ public void testBugIssue464() throws Exception {
 
         String tableName = "testBugIssue464";
 
-        final Statement statement = this.conn.createStatement();
```
brunos-bq, line 12727: NIT: was that unused?

brunos-bq (approved):

src/test/java/testsuite/regression/StatementRegressionTest.java
crystall-bitquill: Yes

src/test/java/testsuite/regression/StatementRegressionTest.java
```diff
@@ -12749,4 +12748,53 @@ public void testBugIssue464() throws Exception {
...
+
+        Statement stmt = conn.createStatement();
+        stmt.addBatch("INSERT INTO " + tableName + " VALUES(1)");
+        stmt.addBatch("INSERT INTO " + tableName + " VALUES(2)");
+
+        try {
+            stmt.executeBatch();
+        } catch (BatchUpdateException e) {
+            fail();
+        }
```
karenc-bq, line 12777: Why do we explicitly catch and fail here instead of adding the exception to the method signature?

karenc-bq (approved):

PR #544, opened 2024-01-26
fix: log message

jasonli-improving:
### Summary

Fix missing log message argument

### Description

Refactored function to pass failing argument into exception, making the log message clearer


sergiyvamz (approved):

aaronchung-bitquill (approved):
Nice!

Issue #579, opened 2024-05-21
Exception thrown when using IAM authentication in CN RDS


fantasy194:
### Describe the bug

An exception was thrown when I enabled IAM authentication on the URL parameter of CN RDS: 
Mon May 20 09:31:55 UTC 2024 TRACE: Unsupported AWS hostname 'xxxxx.cluster-xxxxxxxxxx.rds.cn-northwest-1.amazonaws.com.cn'. Amazon domain name in format *.AWS-Region.rds.amazonaws.com is expected

My config is:
spring.datasource.url=jdbc:mysql:aws://xxxxx.cluster-xxxxxxxxxx.rds.cn-northwest-1.amazonaws.com.cn:3306/xxxxxxxx?useUnicode=true&characterEncoding=UTF-8&useSSL=false&useAwsIam=true

I have tried to set 'clusterInstanceHostPattern' but is not working.

### Expected Behavior

Works normally

### Current Behavior

An exception thrown block the application initialization

### Reproduction Steps

Set a CN RDS URL like '?.rds.AWS-CNRegion.amazonaws.com.cn' and enable IAM authentication

### Possible Solution

com.mysql.cj.protocol.a.authentication.AwsIamAuthenticationTokenHelper.getRdsRegion() uses fix regex expression
![image](https://github.com/awslabs/aws-mysql-jdbc/assets/22437438/9189a8d1-fc6b-4df4-9318-75eaae6fb670)

### The AWS JDBC Driver for MySQL version used

1.1.14

### JDK version used

graalvm ce 11

### Operating System and version

Windows11

aaronchung-bitquill:
Hi @fantasy194 ,

Thanks for reaching out and raising this issue.

We'll take a look at this and keep you updated as we investigate.

Thank you for your patience!

aaronchung-bitquill:
Hi @fantasy194 

While this fix is in progress, I wanted to bring to your attention that the maintenance window for the AWS MySQL JDBC Driver will be ending very soon on July 25 2024 as per the [maintenance policy linked here](https://github.com/awslabs/aws-mysql-jdbc/blob/main/RELEASE_POLICY.md#maintenance-policy). 

I would highly recommend that before the AWS MySQL JDBC Driver maintenance window ends, adopting the [`aws-advanced-jdbc-wrapper`](https://github.com/aws/aws-advanced-jdbc-wrapper). It offers many of the same features and more. And most importantly, will still be covered by its maintenance window. 

[...]

PR #582, opened 2024-05-27
fix: IAM auth in CN RDS (#579)

aaronchung-bitquill:
### Summary

fix: IAM auth in CN RDS (#579)

### Description

Previously `AwsIamAuthenticationTokenHelper#getRegion` aurora DNS pattern did not support CN regions. Added an aurora DNS pattern for CN region specifically.

### DIFF
--- a/src/main/protocol-impl/java/com/mysql/cj/protocol/a/authentication/AwsIamAuthenticationTokenHelper.java
+++ b/src/main/protocol-impl/java/com/mysql/cj/protocol/a/authentication/AwsIamAuthenticationTokenHelper.java
@@ -31,6 +31,7 @@
 
 package com.mysql.cj.protocol.a.authentication;
 
+import com.mysql.cj.jdbc.ha.util.RdsUtils;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.rds.RdsUtilities;
@@ -50,7 +51,7 @@ public class AwsIamAuthenticationTokenHelper {
   private final String hostname;
   private final int port;
   private final Log log;
-  private static final int REGION_MATCHER_GROUP = 3;
+  private static final String REGION_MATCHER_GROUP = "region";
 
   public AwsIamAuthenticationTokenHelper(final String hostname, final int port, final String logger) {
     this.log = LogFactory.getLogger(logger, Log.LOGGER_INSTANCE_NAME);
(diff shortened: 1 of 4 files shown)


src/main/protocol-impl/java/com/mysql/cj/protocol/a/authentication/AwsIamAuthenticationTokenHelper.java
```diff
@@ -85,17 +85,25 @@ private Region getRdsRegion() {
     // Check Hostname
     final Pattern auroraDnsPattern =
         Pattern.compile(
-            "(.+)\\.(proxy-|cluster-|cluster-ro-|cluster-custom-)?[a-zA-Z0-9]+\\.([a-zA-Z0-9\\-]+)\\.rds\\.amazonaws\\.com",
+            "(.+)\\.(proxy-|cluster-|cluster-ro-|cluster-custom-)?[a-zA-Z0-9]+\\.([a-zA-Z0-9\\-]+)\\.rds\\.amazonaws\\.com(\\.cn)?",
```
sergiyvamz, line 88: Why do we need the last (\\.cn) in this regexp? The code checks for China url pattern further.

sergiyvamz: I'd suggest to replocate entire method getRdsRegion() to RdsUtils.

brunos-bq, line 88: i think there might be some confusion here (?)

the first regex was supposed to try to match non-CN endpoints

and then there was the CN-matcher for CN endpoints

do you want to merge them so one regex matches both?

I would suggest to put the regex string as a static variable and in the comments put the patterns that are valid

src/main/protocol-impl/java/com/mysql/cj/protocol/a/authentication/AwsIamAuthenticationTokenHelper.java
```diff
@@ -85,17 +85,25 @@ private Region getRdsRegion() {
...
+      if (!chinaMatcher.find()) {
+        // Does not match Amazon's Hostname, throw exception
+        final String exceptionMessage = Messages.getString(
+            "AuthenticationAwsIamPlugin.UnsupportedHostname",
+            new String[]{hostname});
+
+        log.logTrace(exceptionMessage);
+        throw ExceptionFactory.createException(exceptionMessage);
+      }
+      matcher = chinaMatcher;
```
brunos-bq, line 106: do you need to do something with the matcher here?

src/test/java/testsuite/simple/AwsIamAuthenticationHelperTest.java
```diff
@@ -158,4 +158,15 @@ public void test_6_InvalidRegion() {
             )
         );
     }
+    /**
+     *
+     */
+    @Test
+    public void test_7_ValidChinaHostAndRegion() {
+        Assertions.assertNotNull(new AwsIamAuthenticationTokenHelper(
```
brunos-bq, line 166: maybe you should add a couple of examples that have the `.cn` at the end but have other irregularities in the middle of the URL to be sure your regex is catching what you want

src/main/protocol-impl/java/com/mysql/cj/protocol/a/authentication/AwsIamAuthenticationTokenHelper.java
aaronchung-bitquill: I was double checking this DNS pattern against the one in `RdSUtils.java` and it contains `(\.cn)?` at the end. Same thing in the `aws-advanced-jdbc-wrapper`. So it looks like this pattern may have been missing it.

aaronchung-bitquill: Added `(\\.cn)?` to the end because i noticed that the pattern in RdsUtils.java has it, but it doesnt have it here. 
Personally, I'd prefer not to mix them. I'd rather keep it simple with two separate patterns, than one more complex one. 
I will put them in static vars are you suggested though.

crystall-bitquill (approved):

PR #641, opened 2024-10-23
fix: query cache size version cutoff

GregCKrause:
> The query cache is deprecated as of MySQL 5.7.20, and is removed in MySQL 8.0
> https://dev.mysql.com/doc/refman/5.7/en/query-cache-configuration.html

This pull request lowers the minimum version cutoff for `query_cache_size` and `query_cache_type` from `8.0.3` to `8.0.0`.


davecramer:
This driver has been deprecated. We will not be supporting this. Please use https://github.com/aws/aws-advanced-jdbc-wrapper

